PR #5, opened 2023-02-01
Display icons in tab headers, bump Eto SDK to 2.7.3

jonko0493:


src/SerialLoops/EditorTabsPanel.cs
```diff
@@ -46,12 +48,10 @@ void InitializeComponent()
 
         public static Icon GetItemIcon(ItemDescription.ItemType type)
         {
-            //todo battle with this and figure out loading from a resx?
-            //return Icon.FromResource($"SerialLoops.Icons.{type}.png", Assembly.GetExecutingAssembly());
-            throw new NotImplementedException();
+            return Icon.FromResource($"SerialLoops.Icons.{type}.png", Assembly.GetExecutingAssembly()).WithSize(16, 16);
```
WiIIiam278, line 51: I think you can actually just call `Icon.FromResource($"SerialLoops.Icons.{type}.png").WithSize(16, 16);` here since `FromResource` defaults to the current assembly if one is not passed

WiIIiam278 (approved):

src/SerialLoops/EditorTabsPanel.cs
jonko0493: oh, pog, nice

PR #14, opened 2023-02-02
Add chibi items

jonko0493:
image

~Animation can stutter a little, but that's okay for now at least~ *Looks like this might only be the case on resource-starved machines lol, worked fine on all my other test machines.* Chibis are displayed and there's a dropdown to select which chibi to view.


WiIIiam278 (approved):

WiIIiam278:
I do think the frames would be better along the bottom with a way of pausing/playing the animation at each frame (ideally it'd show which frame you were on, too, kind of like a timeline). Maybe in a second pass? This looks great though :)

jonko0493:
@WiIIiam278 I think we're gonna want to refactor all of these eventually anyway, yeah.

PR #15, opened 2023-02-02
Add tab management context menu

WiIIiam278:
Adds a context menu for closing tabs and fixes doubling up of scroll bars on item list menus
image


jonko0493 (approved):
Beautiful, I already want this

PR #16, opened 2023-02-02
Add character sprite items

jonko0493:
image

Add a single animated image for the character sprites.

### DIFF
--- a/src/SerialLoops/Editors/CharacterSpriteEditor.cs
+++ b/src/SerialLoops/Editors/CharacterSpriteEditor.cs
@@ -1,0 +1,23 @@
+﻿using Eto.Forms;
+using HaruhiChokuretsuLib.Util;
+using SerialLoops.Lib.Items;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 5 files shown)


jonko0493:
The perf isn't great on this one, but loading all the sprites beforehand destroys loading perf. So it has to generate them at sprite open time, which is fine.

src/SerialLoops/Editors/CharacterSpriteEditor.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+        }
+
+        public override Panel GetEditorPanel()
+        {
+            _sprite = (CharacterSpriteItem)Description;
+            _animatedImage = new AnimatedImage(_sprite.GetLipFlapAnimation(_project));
+            _animatedImage.Play();
+
+            return new StackLayout
```
WiIIiam278, line 24: At some point we should probably make a control for this and background items that allows for zooming and centers the image in the panel

WiIIiam278 (approved):

src/SerialLoops/Editors/CharacterSpriteEditor.cs
jonko0493: Agreed

PR #17, opened 2023-02-03
Add script references to chibis, sprites, and backgrounds

jonko0493:
image

What it says on the tin.

### DIFF
--- a/src/SerialLoops/ReferenceDialog.eto.cs
+++ b/src/SerialLoops/ReferenceDialog.eto.cs
@@ -5,6 +5,7 @@ using SerialLoops.Controls;
 using SerialLoops.Lib;
 using SerialLoops.Lib.Items;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace SerialLoops
 {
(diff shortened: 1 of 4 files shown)


src/SerialLoops/ReferenceDialog.eto.cs
```diff
@@ -47,7 +48,32 @@ private void InitializeComponent()
 
         private List<ItemDescription> GetResults()
         {
-            return new List<ItemDescription>(); //todo
+            switch (Item.Type)
```
WiIIiam278, line 51: I think we ought to actually split this into GetReferencesTo() and GetReferencedBy() -- it'll probably be cleaner as a lot of stuff will just have one or the other and not both

WiIIiam278 (changes requested):
might be worth splitting that function? other than that, looks great. this is super useful

src/SerialLoops/ReferenceDialog.eto.cs
jonko0493: Sounds good, will do

WiIIiam278 (approved):

PR #18, opened 2023-02-03
New chibi editor layout

WiIIiam278:
Just a bit of a layout revamp. Also provides room for us to add buttons on the RHS sidebar for doing actual editing.

image

Todo: use a grid style layout for the animation frames, perhaps

### DIFF
--- a/src/SerialLoops/Editors/ChibiEditor.cs
+++ b/src/SerialLoops/Editors/ChibiEditor.cs
@@ -3,14 +3,18 @@ using HaruhiChokuretsuLib.Util;
 using SerialLoops.Lib.Items;
 using SerialLoops.Utility;
 using System;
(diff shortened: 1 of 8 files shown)


src/SerialLoops/Editors/ChibiEditor.cs
```diff
@@ -3,58 +3,103 @@
...
+                {
+                    new GroupBox
+                    {
+                        Text = "Animation",
+                        Padding = 10,
+                        Content = _animationSelection
+                    },
+                    new GroupBox
+                    {
+                        Text = "Pose",
```
jonko0493, line 64: I dunno that these are poses -- more like facing directions, right?

jonko0493 (approved):
Redesign looks fantastic! Thanks so much!

PR #19, opened 2023-02-03
Add scenario item

jonko0493:
image

You can't scroll haha. @WiIIiam278 if you have any ideas about that, lemme know.

We'll want to rework this eventually but for now it's probably fine and let's you view stuff at least.

### DIFF
--- a/src/SerialLoops/Editors/ScenarioEditor.cs
+++ b/src/SerialLoops/Editors/ScenarioEditor.cs
@@ -70,7 +70,7 @@ namespace SerialLoops.Editors
                 tableLayout.Rows.Add(row);
             }
 
-            return tableLayout;
+            return new Scrollable { Content = tableLayout };
         }
 
         private void LinkClick_Click(object sender, System.EventArgs e)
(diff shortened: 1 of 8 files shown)


src/SerialLoops/Editors/ScenarioEditor.cs
```diff
@@ -0,0 +1,86 @@
...
+
+                    default:
+                        row.Cells.Add(new TableCell(new TextBox { Text = parameter }));
+                        break;
+                }
+
+                tableLayout.Rows.Add(row);
+            }
+
+            return tableLayout;
```
WiIIiam278, line 73: To make this scrollable, I think

```cs
return new Scrollable {
    Content = tableLayout
}
```

would _probably_ work?

WiIIiam278 (approved):

src/SerialLoops/Editors/ScenarioEditor.cs
jonko0493: Well that's easy lmao

PR #21, opened 2023-02-04
New script editor!!

jonko0493:
Not quite ready yet, but check it out!


WiIIiam278 (commented):
Gave this a spin.. it's very cool! looks really good, too -- still some stuff to sort out with certain param/command types, but this is a great start! 

I'll lend my UI thoughts: This is a much, much better layout, good idea! I do have a few suggestions, though:
* I think the best thing we can do in terms of UI is remove some of the text by replacing it with the relevant images (e.g. character emotes) and just adding some more icons in general (e.g. file type icons on the relevant select dropdown controls). Perhaps a little icon could be used to represent characters, (kyon, haruhi, etc), too.
* The preview listitems should probably have increased padding and be _slightly_ wider, perhaps? 
* We should probably replace certain dropdowns (like of character sprites) with select menus that can be easily filtered (or look if eto has a searchable dropdown control)
* I also think we should probably change the command parameters (which are currently just taken from your spec) to be more legible (i.e. "spriteEntranceTransition" --> "Sprite Entrance Transition") since they're there for UI and not really for internal use or anything.
* Consider putting each parameter on its own line (less compact, but probably easier to use and familiarize this way I imagine)

Future stuff to consider: 
* With DIALOGUE and SELECT commands I definitely also think having a preview in-built would be an excellent idea, too. This may actually self-address the need to add image previews in so many of the parameter edit controls.
* Some sort of UI for adding, removing and re-ordering commands, which may be hard with a listbox

jonko0493:
* Completely agree on replacing the UI elements with images where possible and adding icons. I'd imagine this would also dovetail with the suggestion for specific selection menus. My main concern for that is caching & perf, but I'm sure we can optimize it to work okay.
* The listbox has a few too many limitations and needs to go and be replaced with something custom &ndash; perhaps reusing sections to some extent? The commands need to be organized into script blocks in order to be able to make sense of them, for one thing, and like you said be able to do add/delete/replace. I actually think something like our existing item tree would work well for that with some tweaks.
* I was planning on having a built-in preview for every command eventually, and I'd _really_ like to have it be cumulative, i.e., the current command accounts for the state defined by previous commands to display what the screen looks like.
* I'll try putting each parameter on its own line for sure.

Additionally, we need to have space for editing the other sections of the event file as well, most notably the properties for the ISO map, the starting chibis, etc.

PR #22, opened 2023-02-07
Implement script sections for script editor

jonko0493:
image

Script editor now has sections listed using an implementation of the section tree control. Additionally, a depth-first search algorithm creates a route through the various sections so the previews are accurate. Finally, I've finished the save function for edited backgrounds.

Edit: added the character sprite preview!


WiIIiam278 (approved):
Glorious work. Looks and feels pretty fantastic imo. Love the sprites in the previews
Maybe still some stuff to do with regards to the list box in terms of repositioning/adding/removing commands and sections, but this is pretty great all things considered.


Misc editor lib comment: in retrospect I just noticed we're still replacing the Japanese with full width English characters. Is that still the best solution to how we're handling text replacement? I wasn't really here for the early Choku ROM hacking so I don't know the details but that seems like something we could improve on

jonko0493:
Yeah, definitely still need to figure out how to handle command insertion/deletion/replacement and potentially reordering. The first three don't seem so hard, but i'm not sure how to do the last off the top of my head.

As for the text replacement, the only other solution would be to somehow get Chokuretsu to recognize half-width characters, which would be a helluva hack since its character parsing routine parsers things two bytes at a time. I prefer the current solution we have, tbh, as it's less heavy on the ROM. That being said, full-width characters are obviously not going to be the way we input text in the application; rather, it will do the same character mapping that we do with the resx files on string import. I haven't gotten working on that yet, but maybe I'll do that next!

WiIIiam278:
> Yeah, definitely still need to figure out how to handle command insertion/deletion/replacement and potentially reordering. The first three don't seem so hard, but i'm not sure how to do the last off the top of my head.

Perhaps some buttons to just "move up" and "move down" when you're click? Or, even better, dragging -- though that may be outside of the remit of a listbox.

I think we'd need to implement some sort of handling/warning to check illegal script ordering and such

PR #23, opened 2023-02-07
Add build capabilities

jonko0493:
Adds build capabilities and a preferences menu for adjusting config. Currently does not build ASM hacks, just the archives.

Still needs to be tested on Mac & Linux before merging.


WiIIiam278 (approved):
My only suggestion was to be the toolbar, but you sorted that. Nice work all round, will look into doing icons for that soon.

My only comment is the static inclusion of e50_newsize.png -- this seems to be the FanTL version, so doesn't this break the jp dialogue? I imagine it may also cause issues where say, another language wishes to add additional nonlatin characters (Korean, maybe). Could we pull this from the assets repo to maintain compatibility with future languages that may need to customize the fontset? Everything else looks great though

jonko0493:
So re: the font -- it does break Japanese characters, but it isn't included if language is set to Japanese. For now, I think we should avoid referencing the Assets repo at all and encourage hackers to use the translated ROM as a base if they want the English language graphics & text by default. Though now that you mention it, the inclusion of the font isn't strictly necessary since it's already included with the translated ROM, so actually now that you mention it I'm just going to dump it entirely lol.

PR #26, opened 2023-02-09
Add app icons

WiIIiam278:
image

may revise these later on but was bothering me enough to have a first pass


jonko0493:
ooh overall i love the design! Should we maybe make the SL more distinct from the spiral? Like by changing one of their colors or by making the spirals transparent, maybe?

WiIIiam278:
Yeah probably. May run through a few more prototypes

WiIIiam278:
Updated with revised design

PR #27, opened 2023-02-10
Add recent projects, build & run, toolbar icons

WiIIiam278:
image


src/SerialLoops.Lib/Config.cs
```diff
@@ -74,10 +73,23 @@ private static Config GetDefault()
...
+            // TODO: Probably make a way of defining "presets" of common emulator install paths on different platforms.
+            // Ideally this should be as painless as possible.
+            string emulatorPath = "";
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                emulatorPath = Path.Combine("/Applications", "melonDS.app");
+            }
+            if (!Directory.Exists(emulatorPath))
+            {
+                emulatorPath = "";
```
jonko0493, line 85: Probably should log a warning here for debugging purposes.

src/SerialLoops.Lib/RecentProjects.cs
```diff
@@ -0,0 +1,70 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.Json;
+using HaruhiChokuretsuLib.Util;
+
+namespace SerialLoops.Lib
```
jonko0493, line 7: Is there a reason for making RecentProjects a separate file rather than a list within the Config file? I feel like we could keep all this code in one place and then have one less object to pass around as well as one less file to keep track of. You could even keep this object for simplicity and then add it as a property of the Config object.

src/SerialLoops/MainForm.eto.cs
```diff
@@ -195,6 +241,41 @@ private async void BuildBaseProject_Executed(object sender, EventArgs e)
...
             }
         }
+        
+        private async void BuildAndRunProject_Executed(object sender, EventArgs e)
+        {
+            if (OpenProject is not null)
+            {
+                if (CurrentConfig.EmulatorPath is null)
+                {
+                    MessageBox.Show("No emulator path set. Please set the path to your emulator.", "No Emulator Path", MessageBoxType.Warning);
```
jonko0493, line 251: Perhaps consider logging a warning here.

src/SerialLoops/MainForm.eto.cs
```diff
@@ -195,6 +241,41 @@ private async void BuildBaseProject_Executed(object sender, EventArgs e)
...
+                        if (emulatorExecutable.EndsWith(".app"))
+                        {
+                            emulatorExecutable = Path.Combine(CurrentConfig.EmulatorPath, "Contents", "MacOS", Path.GetFileNameWithoutExtension(CurrentConfig.EmulatorPath));
+                        }
+                        
+                        Process.Start(emulatorExecutable, $"\"{Path.Combine(OpenProject.MainDirectory, $"{OpenProject.Name}.nds")}\"");
+                    } catch (Exception ex)
+                    {
+                        _log.LogError($"Failed to start emulator: {ex.Message}");
+                        MessageBox.Show("Failed to start your emulator", "Emulator Startup Failed", MessageBoxType.Error);
```
jonko0493, line 270: This will open two message boxes when using the default LoopyLogger (which opens a message box when logging an error).

jonko0493 (approved):
Looks great! All of this stuff will make my life easier haha

src/SerialLoops.Lib/RecentProjects.cs
WiIIiam278: Yeah, that would've been easier. Main reasons for doing this was that I figured this was more of a "cache" than it was config properties, and I figured if someone wants to edit their config through the file itself (fairly reasonable thing to want to do), that file should probably not hold cached data.

src/SerialLoops/MainForm.eto.cs
WiIIiam278: oh, neat, okay.

src/SerialLoops.Lib/RecentProjects.cs
jonko0493: Understandable; feel free to keep as is then.

PR #28, opened 2023-02-10
Add BGM editor

jonko0493:
Draft for now because there's still some stuff I need to sort out (~namely, the BGM names from the extras file are wrong bc the indices are not related to filename but to the index in SND_DS.S~ nope, didn't have to reverse it, it's just the order of the files lol). I may also add voice file editors before this is merged (and depending on how ambitious I get, file replacement).


src/SerialLoops/Utility/WaveProviderStream.cs
```diff
@@ -0,0 +1,31 @@
+﻿using NAudio.Wave;
+using System;
+
+namespace SerialLoops.Utility
+{
+    public class WaveProviderStream : WaveStream
```
WiIIiam278, line 6: Any reason this class is exactly duplicated in both SerialLoops.Lib and SerialLoops? Can't SerialLoops just use the SerialLoops.Lib impl?

WiIIiam278 (approved):
Super cool stuff in general

src/SerialLoops/Utility/WaveProviderStream.cs
jonko0493: OOPS i thought i deleted it

jonko0493: thanks for the catch!

src/SerialLoops/Controls/SoundPlayerPanel.cs
```diff
@@ -0,0 +1,78 @@
...
+            _player.Stop();
+        }
+
+        private void PlayPauseButton_Click(object sender, System.EventArgs e)
+        {
+            Button playPauseButton = (Button)sender;
+            if (_player.IsPlaying)
+            {
+                _player.Pause();
+                playPauseButton.Text = "▶️";
```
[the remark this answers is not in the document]
jonko0493: Yes agreed! I knew it would, but I'm not good at icons and such, so I was presumptuously leaving that to you :)

src/SerialLoops.Lib/Util/WaveProviderStream.cs
```diff
@@ -0,0 +1,31 @@
...
+            _source = source;
+        }
+
+        public override WaveFormat WaveFormat => _source.WaveFormat;
+
+        public override bool CanSeek => false;
+
+        public override long Length => int.MaxValue;
+
+        public override long Position { get => position; set => throw new NotImplementedException(); }
```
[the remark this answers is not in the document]
jonko0493: We actually _can_ do arbitrary seeking now! This guy isn't the final stream; we read him out fully into a memory that becomes the final stream.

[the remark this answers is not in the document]
jonko0493: And that stream is seekable.

src/SerialLoops/Utility/ControlGenerator.cs
```diff
@@ -35,8 +35,11 @@ public static TableLayout GetControlWithLabelTable(string title, Control control
         
         public static StackLayout GetFileLink(ItemDescription description, EditorTabsPanel editorTabs, ILogger log)
         {
-            LinkButton link = new() { Text = description.Name };
-            link.Click += (s, e) => { editorTabs.OpenTab(description, log); };
+            ClearableLinkButton link = new() { Text = description.Name };
+            if (description.Name != "NONE")
```
WiIIiam278, line 39: We should probably set link to be a text control rather than a link for NONE values and return that in the stack layout below

jonko0493: that's smart. i'll try doing that

jonko0493: I tried doing this -- problem is some of our other logic relies on their being a `ClearableLinkButton` there so the text of it can be modified. I think maybe we can do something with either restyling it or by passing the StackLayoutItem and clearing its items every time we modify a link, but I'm not sure which is best.

WiIIiam278: Ah yeah, before I was just updating the contents of the parent controls with a new LinkButton -- but this is probably smarter. No biggie either way.

PR #30, opened 2023-02-13
Add topic editor

jonko0493:
Adds topic items and a topic editor. Also adds references to topics and scripts accordingly.

image

Doesn't look amazing but is functional.


src/SerialLoops/Editors/TopicEditor.cs
```diff
@@ -0,0 +1,42 @@
...
+            TextBox titleTextBox = new() { Text = _topic.Topic.Title, Width = 200 };
+
+            DropDown linkedScriptDropDown = new();
+            linkedScriptDropDown.Items.AddRange(_project.Items.Where(i => i.Type == ItemDescription.ItemType.Script).Select(s => new ListItem { Key = s.Name, Text = s.Name }));
+            linkedScriptDropDown.SelectedKey = _project.Items.First(i => i.Type == ItemDescription.ItemType.Script && ((ScriptItem)i).Event.Index == _topic.Topic.EventIndex).Name;
+
+            StackLayout shortsLayout = new();
+            for (int i = 0; i < _topic.Topic.UnknownShorts.Length; i++)
+            {
+                shortsLayout.Items.Add(ControlGenerator.GetControlWithLabel($"Short {i}", new TextBox() { Text = _topic.Topic.UnknownShorts[i].ToString() }));
```
WiIIiam278, line 33: Do we know what these do yet out of interest?

src/SerialLoops/Editors/TopicEditor.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+
+            return new TableLayout(ControlGenerator.GetControlWithLabelTable("ID", idLabel),
+                ControlGenerator.GetControlWithLabel("Title", titleTextBox),
+                ControlGenerator.GetControlWithLabel("Associated Script", linkedScriptDropDown),
```
WiIIiam278, line 38: Could we perhaps do an iconified blue file link via the ControlGenerator next to the dropdown?
Given how many scripts there are we should probably make this a filterable selection modal in the future

WiIIiam278 (approved):
Most excellent :)

src/SerialLoops.Lib/Items/TopicItem.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Linq;
+
+namespace SerialLoops.Lib.Items
+{
+    public class TopicItem : Item
+    {
+        public TopicStruct Topic { get; set; }
+        public (string ScriptName, ScriptCommandInvocation command)[] ScriptUses { get; set; }
+
+        public TopicItem(TopicStruct topicStruct, Project project) : base($"{topicStruct.Id} {topicStruct.Title}", ItemType.Topic)
```
WiIIiam278, line 11: Just a thought -- this makes the item name modifiable, which probably isn't a problem as the fact there's a proceeding ID still makes them unique, but may be a bit confusing and cause issues at some point when users expect certain things of item names.
Perhaps we may want to consider some sort of "item display name" separate from the actual item "id" we're assigning each one?

src/SerialLoops/Editors/TopicEditor.cs
jonko0493: No :") I'm gonna work on it at some point but they're pretty confusing and a bit low priority at the moment.

src/SerialLoops.Lib/Items/TopicItem.cs
jonko0493: Oh you're right. I actually did introduce an item display name for this purpose; I'll see if I can use that.

src/SerialLoops/Editors/TopicEditor.cs
jonko0493: Yes, I will add this! And agreed, we're going to want filterable modals at some points.

WiIIiam278 (commented):
Could we perhaps update the topic links in the PuzzleEditor to be item links instead of the placeholder links with the topic ID text, too?

src/SerialLoops/Editors/PuzzleEditor.cs
```diff
@@ -31,10 +33,16 @@ public override Container GetEditorPanel()
             };
 
             GroupBox topicsBox = new() { Text = "Associated Main Topics", Padding = 5 };
-            StackLayout topics = new() { Orientation = Orientation.Horizontal, Spacing = 5 };
-            foreach (var (topic, _) in _puzzle.Puzzle.AssociatedTopics.Take(_puzzle.Puzzle.AssociatedTopics.Count - 1))
+            StackLayout topics = new() { Orientation = Orientation.Vertical, Spacing = 5 };
```
WiIIiam278, line 36: This change seems to cause the GroupBox name to not display properly on mac as the name is too long for the now thinner box

image

src/SerialLoops.Lib/Items/ItemDescription.cs
```diff
@@ -4,14 +4,23 @@ public class ItemDescription
     {
 
         public string Name { get; protected set; }
-        public string DisplayName => UnsavedChanges ? $"{Name} *" : Name;
+        public string DisplayName { get; protected set; }
```
WiIIiam278, line 7: We might want to have Search use DisplayName instead of name to show suggestions for topic names as right now it's just using Name

WiIIiam278 (changes requested):
Just a fix needed for item selection context menus as a result of the DisplayName change

src/SerialLoops.Lib/Items/ItemDescription.cs
jonko0493: Done!

src/SerialLoops/Controls/ItemExplorerPanel.cs
```diff
@@ -24,7 +24,7 @@ protected override void ItemList_ItemClicked(object sender, EventArgs e)
         {
             if (sender is SectionListTreeGridView view)
             {
-                ItemDescription item = _project.FindItem(view.SelectedItem?.Text);
+                ItemDescription item = _project.FindItem(view.SelectedItem?.Text.Split(" - ")[0]);
```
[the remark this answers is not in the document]
jonko0493: Refactored this into `FindItem`. As for ID - Nice name -- we'll do that through the current system of Name/DisplayName and not have any breakage. :)

PR #33, opened 2023-02-13
Progress tracking on loading item tabs

WiIIiam278:
This works, but is flawed: Because an `Editor` extends `DocumentPanel`, we _must_ instantiate it on the UI thread. Given that all logic for loading items is tangled up in the actual constructors of control objects, the result is that the progress bar is only able to update between breaks in main thread processing block calls. In the future, this may be worth another refactor. 

It does work, though, at the very least, and means it doesn't look like the application has crashed every time you load an audio item -- it's just that the bar isn't able to be that accurate when loading BGMs, for instance.

image

### DIFF
--- a/src/SerialLoops/Controls/EditorTabsPanel.cs
+++ b/src/SerialLoops/Controls/EditorTabsPanel.cs
@@ -1,8 +1,11 @@
 ﻿using Eto.Forms;
+using HaruhiChokuretsuLib.Archive.Event;
 using HaruhiChokuretsuLib.Util;
 using SerialLoops.Editors;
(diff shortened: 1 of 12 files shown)


jonko0493:
Since #34 eliminates loading times for BGM & voice items, I'm not sure this is still worth introducing (especially since the logic remains on the UI thread). Lemme know what you think.

WiIIiam278:
> Since #34 eliminates loading times for BGM & voice items, I'm not sure this is still worth introducing (especially since the logic remains on the UI thread). Lemme know what you think.

Yeah ok--if that's now instantaneous there's not much point :)

PR #34, opened 2023-02-14
Use cross-plat IWavePlayer to eliminate loading times for sound players

jonko0493:
This structurally changes the project a bit as we needed the ability to add native code. Everything else works the same.

Notable limitation: on Windows, you can play as many sounds at the same time as you like. However, on Mac and Linux, you can only play one sound at a time.

Also slipped in a couple of fixes to the script preview renderer, namely:
* Added `SET_READ_FLAG` as a command that prohibits graph edges (possibly need to rename that command to `BACK`), which fixes rendering in topic files and a few other places
* Sprite exits are handled _after_ dialogue is advanced, so now the sprite renderer handles those on the following command


WiIIiam278 (approved):
Looks really good, works a treat on Mac. Did notice that voiced line references don't populate quite right on this for some reason, but that's probably unrelated.

PR #35, opened 2023-02-15
Add chibi emotes preview, group selection editor, and saving to scenario editor

jonko0493:
image

The chibi emotes themselves have a non-transparent background for reasons I don't yet fully understand... however, for the purposes of a preview it's probably fine for now. I wanted to render the bubble as well but that seems to be somewhat difficult.

image

Group selections are not yet editable, but they can be viewed. 

image

Scenario file can now actually be edited, including adding and removing commands.


WiIIiam278 (approved):
Looks good to me for now! We may want to change the layout in the future to avoid the page being too long (maybe a drop down for selecting which route set to view)

jonko0493:
Agreed -- that would also improve load times. I'll save it for the future work I'm gonna do on it anyway to make it editable.

PR #39, opened 2023-02-19
Add complete parameter editing to the script editor

jonko0493:
Adds the ability to edit all script command parameters (except for the chess-related parameters which are still poorly-understood).

Also adds some other neat things:
* A "searchable text" field to make searching through files easier -- might redo how this is done at some point but was useful
* Bugfix for the NONE selection crash
* Fixes a macOS context menu crash (probably)
* Adds Dialogue Properties items
* Adds the background palette effects to the preview (with color matrices!!)
* Adds a "background export" button to the background editor, allowing for exporting images
* Adds new RE info to the topic & group selection editors
* Adds half-width text replacement to most files (TODO: conditionalize it on config)
* Adds auto-nice quotes to text editors
* Fixup tab context menus to work on tabs


src/SerialLoops/Editors/ScriptEditor.cs
```diff
@@ -807,46 +984,405 @@ private void ChibiDropDown_SelectedKeyChanged(object sender, EventArgs e)
...
+        private void SpriteEntranceDropDown_SelectedKeyChanged(object sender, EventArgs e)
+        {
+            ScriptCommandDropDown dropDown = (ScriptCommandDropDown)sender;
+            _log.Log($"Attempting to modify parameter {dropDown.ParameterIndex} to sprite entrance {dropDown.SelectedKey} in {dropDown.Command.Index} in file {_script.Name}...");
+            ((SpriteEntranceScriptParameter)dropDown.Command.Parameters[dropDown.ParameterIndex]).EntranceTransition =
+                Enum.Parse<SpriteEntranceScriptParameter.SpriteEntranceTransition>(dropDown.SelectedKey);
+            _script.Event.ScriptSections[_script.Event.ScriptSections.IndexOf(dropDown.Command.Section)]
+                .Objects[dropDown.Command.Index].Parameters[dropDown.ParameterIndex] =
+                (short)Enum.Parse<SpriteEntranceScriptParameter.SpriteEntranceTransition>(dropDown.SelectedKey);
+            UpdateTabTitle(false);
```
WiIIiam278, line 1257: This needs to update the preview too, as changing the transition from left-right will cause character sprites to be re-positioned (though this isn't the case with the exit transition dropdown as the effect of that is only apparent to the subsequent command).
```suggestion
            UpdateTabTitle(false);
            Application.Instance.Invoke(() => UpdatePreview());
```

src/SerialLoops/Editors/ScriptEditor.cs
```diff
@@ -368,18 +435,21 @@ private void CommandsPanel_SelectedIndexChanged(object sender, EventArgs e)
...
                         sfxModeDropDown.SelectedKey = ((SfxModeScriptParameter)parameter).Mode.ToString();
+                        sfxModeDropDown.SelectedKeyChanged += SfxModeDropDown_SelectedKeyChanged;
 
                         ((TableLayout)controlsTable.Rows.Last().Cells[0].Control).Rows[0].Cells.Add(
                             ControlGenerator.GetControlWithLabel(parameter.Name, sfxModeDropDown));
                         break;
 
                     case ScriptParameter.ParameterType.SHORT:
+                        ScriptCommandTextBox shortTextBox = new() { Command = command, ParameterIndex = i, Text = ((ShortScriptParameter)parameter).Value.ToString() };
+                        shortTextBox.TextChanged += ShortTextBox_TextChanged;
```
WiIIiam278, line 449: In the case of an update to the Sprite Layer value on dialog boxes, we ought to trigger a redraw of the preview to update how the layers render, so may be worth adding a special case for that box

src/SerialLoops/Editors/ScriptEditor.cs
```diff
@@ -368,18 +435,21 @@ private void CommandsPanel_SelectedIndexChanged(object sender, EventArgs e)
...
+                        sfxModeDropDown.Items.AddRange(Enum.GetNames<SfxModeScriptParameter.SfxMode>().Select(t => new ListItem { Text = t, Key = t }));
                         sfxModeDropDown.SelectedKey = ((SfxModeScriptParameter)parameter).Mode.ToString();
+                        sfxModeDropDown.SelectedKeyChanged += SfxModeDropDown_SelectedKeyChanged;
 
                         ((TableLayout)controlsTable.Rows.Last().Cells[0].Control).Rows[0].Cells.Add(
                             ControlGenerator.GetControlWithLabel(parameter.Name, sfxModeDropDown));
                         break;
 
                     case ScriptParameter.ParameterType.SHORT:
+                        ScriptCommandTextBox shortTextBox = new() { Command = command, ParameterIndex = i, Text = ((ShortScriptParameter)parameter).Value.ToString() };
```
WiIIiam278, line 448: We may want to consider using a [`NumericStepper`](http://pages.picoe.ca/docs/api/html/T_Eto_Forms_NumericStepper.htm) for controls like this which has nicer platform support for restricting values to just numbers. Plus, we can specify min and max values that way

src/SerialLoops/Editors/ScriptEditor.cs
```diff
@@ -807,46 +984,405 @@ private void ChibiDropDown_SelectedKeyChanged(object sender, EventArgs e)
...
+            TopicSelectButton button = (TopicSelectButton)sender;
+
+            ScriptCommandDropDown topicDropDown = new() { Command = button.ScriptCommand, ParameterIndex = button.ParameterIndex };
+            topicDropDown.Items.AddRange(_project.Items.Where(i => i.Type == ItemDescription.ItemType.Topic)
+                .Select(t => new ListItem { Key = t.DisplayName, Text = t.DisplayName }));
+            topicDropDown.SelectedIndex = 0;
+            topicDropDown.SelectedIndexChanged += TopicDropDown_SelectedIndexChanged;
+
+            StackLayout topicLink = ControlGenerator.GetFileLink(_project.Items.FirstOrDefault(i => i.Type == ItemDescription.ItemType.Topic), _tabs, _log);
+            topicDropDown.Link = (ClearableLinkButton)topicLink.Items[1].Control;
```
WiIIiam278, line 1343: I got this to crash for me once, though I can't reproduce. I think that's because I found something that referenced a topic that no longer exists, maybe? Or perhaps this logic depends on the display names of the items, still, maybe? The crash was caused by a NullReferenceException when resolving the topic item to open

src/SerialLoops.Lib/Items/ScriptItem.cs
```diff
@@ -22,6 +22,11 @@ public ScriptItem(EventFile evt) : base(evt.Name[0..^1], ItemType.Script)
             Event = evt;
 
             Graph.AddVertexRange(Event.ScriptSections);
+
+            SearchableText = string.Join('\n', evt.ScriptSections.SelectMany(s => s.Objects.Select(c => c.Command.Mnemonic))
+                .Concat(evt.ConditionalsSection.Objects));
+                //.Concat(evt.LabelsSection.Objects.Select(l => l.Name))
```
WiIIiam278, line 28: I take it adding these values to the searchable text causes issues still atm?

src/SerialLoops.Lib/Util/IProgressTracker.cs
```diff
@@ -10,7 +10,8 @@ public interface IProgressTracker
 
         public void Focus(string item, int count)
         {
-            Total += count;
+            Total = count;
```
WiIIiam278, line 13: WOW how did I not notice this earlier?! lol.

src/SerialLoops/Editors/DialogueConfigEditor.cs
```diff
@@ -0,0 +1,42 @@
...
+            characterDropDown.SelectedKey = _dialogueConfig.DialogueConfig.Character.ToString();
+
+            return new StackLayout
+            {
+                Spacing = 5,
+                Items =
+                {
+                    ControlGenerator.GetControlWithLabel("Character", characterDropDown),
+                    ControlGenerator.GetControlWithLabel("Voice Font", new TextBox { Text = _dialogueConfig.DialogueConfig.VoiceFont.ToString() }),
+                    ControlGenerator.GetControlWithLabel("Text Timer", new TextBox { Text = _dialogueConfig.DialogueConfig.TextTimer.ToString() }),
```
WiIIiam278, line 37: Perhaps NumericSteppers would be better here, too?
```suggestion
                    ControlGenerator.GetControlWithLabel("Voice Font", new NumericStepper { Value = _dialogueConfig.DialogueConfig.VoiceFont }),
                    ControlGenerator.GetControlWithLabel("Text Timer", new NumericStepper { Value = _dialogueConfig.DialogueConfig.TextTimer }),
```

src/SerialLoops/Editors/ScriptEditor.cs
```diff
@@ -807,46 +984,405 @@ private void ChibiDropDown_SelectedKeyChanged(object sender, EventArgs e)
...
+                Enum.Parse<ColorMonochromeScriptParameter.ColorMonochrome>(dropDown.SelectedKey);
+            _script.Event.ScriptSections[_script.Event.ScriptSections.IndexOf(dropDown.Command.Section)]
+                .Objects[dropDown.Command.Index].Parameters[dropDown.ParameterIndex] =
+                (short)Enum.Parse<ColorMonochromeScriptParameter.ColorMonochrome>(dropDown.SelectedKey);
+            UpdateTabTitle(false);
+            Application.Instance.Invoke(() => UpdatePreview());
+        }
+        private void ConditionalBox_TextChanged(object sender, EventArgs e)
+        {
+            ScriptCommandTextBox textBox = (ScriptCommandTextBox)sender;
```
WiIIiam278, line 1038: Not sure what the full set of conditionals is, but we should possibly consider doing a UI for this (i.e. auto-completing[?] boxes for variable names, dropdown for different conditional operators) in the future

WiIIiam278 (approved):
Thoroughly awesome work!!

src/SerialLoops.Lib/Items/ScriptItem.cs
jonko0493: Nah I just didn't want to search them haha. Eventually, I want to have like a dropdown in the search dialog that lets you pick what things you want to search.

src/SerialLoops.Lib/Util/IProgressTracker.cs
jonko0493: LOL yeah I was wondering why it wasn't starting at the beginning.

src/SerialLoops/Editors/DialogueConfigEditor.cs
jonko0493: Oooh yes! That's perfect.

src/SerialLoops/Editors/ScriptEditor.cs
jonko0493: I will switch to using those here before merging.

src/SerialLoops/Editors/ScriptEditor.cs
jonko0493: Good point. Yeah, there's definitely going to be cases where we have to specify that it should be redrawn.

src/SerialLoops/Editors/ScriptEditor.cs
jonko0493: I agree that there should be better UX here, but I wasn't sure exactly how to do it with the full range of what can be specified. At the very least, we should have something that specifies what the options are.

src/SerialLoops/Editors/ScriptEditor.cs
jonko0493: Hmmm this shouldn't happen, but I did realize that there's a separate bug here with the display name...some topics have the same name, so that might cause errors.

PR #41, opened 2023-02-22
Adjust chibi editor, add orphan item dialog, add bgm names to DisplayName, add child item counts in tree

WiIIiam278:
* Chibi editor sprite switcher is now on bottom of panel in a new TableRow (because it was causing scaling issues before)
* Added BGM track names to their display name where there is one
* Added a dialog for viewing orphan items (those which are not referenced by any others)
* Removed "Referenced By" because it was confusing and probably not going to be helpful.
* Item browser now displays the number of child items next to the collapsible item row


jonko0493 (approved):
Looks great -- very excited to try this out :)

PR #44, opened 2023-03-02
Add starting chibi selector to script editors

jonko0493:
image
image
image

I've changed my mind on the interactable objects/starting chibis items -- I think it actually might make sense to wait on having map items set up. So this just sets up the starting chibis editor to the script editor. Works! Yay!

### DIFF
--- a/src/SerialLoops/Editors/ScriptEditor.cs
+++ b/src/SerialLoops/Editors/ScriptEditor.cs
@@ -28,7 +28,8 @@ namespace SerialLoops.Editors
         private Dictionary<ScriptSection, List<ScriptItemCommand>> _commands = new();
 
         private TableLayout _detailsLayout = new();
-        private StackLayout _preview = new() { Items = { new SKGuiImage(new(256, 384)) } };
+        private readonly StackLayout _preview = new() { Items = { new SKGuiImage(new(256, 384)) } };
+        private StackLayout _scriptProperties = new();
         private StackLayout _editorControls = new();
         private ScriptCommandListPanel _commandsPanel;
         private CancellationTokenSource _dialogueCancellation, _optionCancellation;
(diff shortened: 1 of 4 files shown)


src/SerialLoops/Editors/ScriptEditor.cs
```diff
@@ -636,6 +746,10 @@ private void UpdatePreview()
             canvas.DrawColor(SKColors.Black);
 
             ScriptItemCommand currentCommand = ((ScriptCommandSectionEntry)_commandsPanel.Viewer.SelectedItem).Command;
+            if (currentCommand is null)
+            {
+                return;
```
WiIIiam278, line 751: afaik this prevents the black empty panel from rendering? Idk, maybe just me, but I feel like it's kinda more intuitive to render a back screen (as though the script hasn't faded in yet) rather than nothing at all and have the properties be floating. Unless keeping that breaks stuff ofc, which, fair enough.

src/SerialLoops/Editors/ScriptEditor.cs
```diff
@@ -87,6 +94,109 @@ private Container GetCommandsContainer()
...
+            availableChibisBox.MouseDoubleClick += (o, args) =>
+            {
+                IListItem chibiSelected = (IListItem)availableChibisBox.SelectedValue;
+                availableChibisBox.Items.Remove(chibiSelected);
+                usedChibisBox.Items.Add(chibiSelected);
+                _script.Event.StartingChibisSection.Objects.Insert(_script.Event.StartingChibisSection.Objects.Count - 1, new() { ChibiIndex = (short)allChibis.First(c => c.DisplayName == chibiSelected.Text).ChibiIndex });
+                UpdateTabTitle(false);
+                Application.Instance.Invoke(() => UpdatePreview());
+            };
+            usedChibisBox.MouseDoubleClick += (o, args) =>
```
WiIIiam278, line 136: maybe cursor events (Left/right) may be good here? Double click is a bit of an odd one -- but this is probably fine :)

WiIIiam278 (approved):

src/SerialLoops/Editors/ScriptEditor.cs
jonko0493: Hmm, yeah we can probably render just a black screen. The problem I was avoiding was the fact that the command list is null when a command isn't selected, but I can fix that I suppose!

src/SerialLoops/Editors/ScriptEditor.cs
jonko0493: I think double click is pretty intuitive -- I think adding drag/drop eventually would be great, but double click makes sense for adding or removing items between the two boxes!

PR #46, opened 2023-03-11
Fixup mac chibi drag & drop bug

jonko0493:
Makes chibi drag/drop exclusive which prevents crashes on mac when two chibis were highlighted simultaneously.


WiIIiam278 (approved):
Just tested on mac -- works great!

PR #49, opened 2023-03-19
Script command reordering and deletion, other misc editor improvements

jonko0493:
The big thing this PR adds is the backend functionality for reordering and deleting commands -- fully works and tested in-game.

Other things added:
* Links to the maps in the puzzle editor
* Rendering (but not editing) the starting camera point on the map
* Sprites are now removed on fade out which is more accurate to in-game behavior


src/SerialLoops/Editors/PuzzleEditor.cs
```diff
@@ -33,6 +33,11 @@ public override Container GetEditorPanel()
                 Spacing = 10,
             };
 
+            MapItem map = (MapItem)_project.Items.First(i => i.Type == ItemDescription.ItemType.Map && i.Name ==
+                _project.Dat.Files.First(f => f.Name == "QMAPS").CastTo<QMapFile>().QMaps[_puzzle.Puzzle.Settings.MapId].Name[0..^2]);
+            LinkButton mapLink = new() { Text = map.Name };
```
WiIIiam278, line 38: perhaps use the ControlGenerator's iconified map link? Maybe also for the main topics -- it's nice to have that be consistent and gives a good clear visual indicator of what the user is going to see when they click it

src/SerialLoops.Lib/Items/MapItem.cs
```diff
@@ -85,6 +87,20 @@ public SKBitmap GetMapImage(ArchiveFile<GraphicsFile> grp, bool displayPathingMa
                 }
             }
 
+            if (displayMapStart)
+            {
+                if (Map.Settings.SlgMode)
+                {
+                    SKPoint start = new(gridZero.X - Map.Settings.StartingPosition.x * 32 + Map.Settings.StartingPosition.y * 32, gridZero.Y + Map.Settings.StartingPosition.x * 16 + Map.Settings.StartingPosition.y * 16 + 16);
+                    canvas.DrawCircle(start, 3, new SKPaint() { Color = SKColors.Pink });
```
WiIIiam278, line 95: Maybe we could consider making this a little film camera icon later on, just so it's obvious what this represents -- I can see this getting a bit lost when we add more UI and stuff to these map scenes

WiIIiam278 (approved):
Nice stuff -- interesting that cam position is set separately to the player character position, but makes sense since they can be controlled independently I suppose.

I'm getting a crash atm when repositioning a script section to the top of a file -- this might be a hangover of some of my work? I only noticed it now.

Other than that though, nice stuff!!

src/SerialLoops.Lib/Items/MapItem.cs
jonko0493: Completely agree -- that would be fantastic, actually! And if it's an icon, we could make it so you could drag it around or something.

jonko0493:
The crash is likely because we need to handle repositioning script _sections_ differently than script _commands_ and it looks like there isn't a distinction being made there atm? which is odd lol

jonko0493:
Oh gotcha, you had a good distinction for moving script sections and I was not respecting it. Lemme see if I can fix that.

PR #53, opened 2023-03-29
Add new script-like scenario editor workflow, scenario clearing

WiIIiam278:
This PR adds a new workflow for the scenario editor to match the script editor for better consistency and ease of use and to address performance issues with the previous table-style scenario editor.

image

This PR also introduces a new 🚀 (Clear Scenario) option.

image

Pressing OK to use this option will clear out the scenario of all commands to let users start planning their game from a blank canvas.

image


src/SerialLoops/Controls/ScenarioCommandListPanel.cs
```diff
@@ -0,0 +1,53 @@
...
+            get => _commands; 
+            set
+            {
+                _commands = value;
+                Viewer?.Items.Clear();
+                _commands.ForEach(c => Viewer?.Items.Add($"{c.Item1} {c.Item2}"));
+            }
+        }
+        public ListBox Viewer { get; private set; }
+        public (ScenarioVerb, string)? SelectedCommand { get => Viewer.SelectedIndex != -1 ? _commands[Viewer.SelectedIndex] : null; }
```
jonko0493, line 23: I think it would be better to give these names like `(ScenarioVerb Verb, string Parameter)` so that further down you can reference them by name rather than `Item1` and `Item2`

jonko0493 (approved):
Looks great!

src/SerialLoops/Editors/ScenarioEditor.cs
```diff
@@ -209,30 +251,55 @@ private void ParameterBox_TextChanged(object sender, EventArgs e)
...
+            _scenario.ScenarioCommands.RemoveAt(selectedIndex);
 
-            _scrollable.Content = GetTable();
+            RefreshCommands();
             UpdateTabTitle(false);
         }
+        
+        private void ClearButton_Click(object sender, EventArgs e)
+        {
+            DialogResult result = MessageBox.Show("Clear all commands from the game scenario?\nThis action is irreversible.", MessageBoxButtons.OKCancel);
```
jonko0493, line 283: You should be able to specify a title in one of the `MessageBox.Show` overloads so the pop-up doesn't just say "Error"

PR #75, opened 2023-04-02
Fixup a few cross-plat bugs

### LINKED ISSUE #62: Numeric steppers & textboxes lose focus as you type
To repro: load up SerialLoops.Mac and attempt to type something into a numeric stepper (for example, create a `SCREEN_FADEIN` command and attempt to adjust the fade time property). The textbox will lose focus as soon as anything is typed. I don't currently know why this happens; it will require investigation.

### LINKED ISSUE #63: Attempting to type in the dialogue box causes a crash
Attempting to type in the dialogue box results in a beach ball followed by a crash. Figuring this one out will also require investigation, but the exception is a stack overflow on the TextArea's regex replace [here](https://github.com/haroohie-club/SerialLoops/blob/9bd04dbf413361a7c882ab68c762b82d27f25e8d/src/SerialLoops/Editors/ScriptEditor.cs#L1993).

### LINKED ISSUE #64: The selected item on the command tree is cleared on Mac after adding commands/sections
Expected behavior: after adding a new command/section, that item should be selected on the tree view.

Actual behavior: after adding a new command/section, the current item is deselected entirely.

This will cause crashes as when nothing is selected in the command tree view, attempting to add a section will cause a crash. This doesn't happen on Windows and Linux (as they always keeps something selected).

### LINKED ISSUE #68: Deleting the last section or command in the tree causes a crash on Linux
Stack trace:

```
 ---> System.NullReferenceException: Object reference not set to an instance of an object.

   at SerialLoops.Controls.ScriptCommandListPanel.<InitializeComponent>b__13_1(Object o, EventArgs e) in /home/jonko/code/SerialLoops/src/SerialLoops/Controls/ScriptCommandListPanel.cs:line 94

   at SerialLoops.Controls.ScriptCommandSectionTreeGridView.DeleteItem(ScriptCommandSectionTreeItem item) in /home/jonko/code/SerialLoops/src/SerialLoops/Controls/ScriptCommandSectionTree.cs:line 227

   at SerialLoops.Editors.ScriptEditor.<>c__DisplayClass20_0.<GetEditorButtons>b__2(Object sender, EventArgs args) in /home/jonko/code/SerialLoops/src/SerialLoops/Editors/ScriptEditor.cs:line 334

   at Eto.PropertyStore.TriggerEvent[T](Object key, Object sender, T args)

   at Eto.Forms.Button.OnClick(EventArgs e)

   at Eto.Forms.Button.Callback.OnClick(Button widget, EventArgs e)

   at Eto.GtkSharp.Forms.Controls.ButtonHandler`3.ButtonConnector.HandleClicked(Object sender, EventArgs e)
```

### LINKED ISSUE #70: Script command selection box does not auto-focus on Mac
It autofocuses on windows, which means that you can hit new command and start typing to select your command. We should be able to do this on Mac as well.

jonko0493:
Fixes #62. Fixes #63. Fixes #64. Fixes #68. Fixes #70.

Quick bug fixes for errors; everything is documented in the linked issues. Tested on macOS and Ubuntu.

### DIFF
--- a/src/SerialLoops/Editors/Editor.cs
+++ b/src/SerialLoops/Editors/Editor.cs
@@ -33,10 +33,11 @@ namespace SerialLoops.Editors
 
         public abstract Container GetEditorPanel();
         
-        public void UpdateTabTitle(bool saved)
+        public void UpdateTabTitle(bool saved, Control caller = null)
         {
             Description.UnsavedChanges = !saved;
             Text = Description.DisplayNameWithStatus;
+            caller?.Focus();
         }
     }
 }
--- a/src/SerialLoops/Editors/Editor.cs
+++ b/src/SerialLoops/Editors/Editor.cs
@@ -32,12 +32,26 @@ namespace SerialLoops.Editors
         }
 
         public abstract Container GetEditorPanel();
-        
+
         public void UpdateTabTitle(bool saved, Control caller = null)
         {
+            int caretIndex = 0;
+            if (caller is not null && caller is TextBox textBox)
+            {
+                // This hack is necessary because macOS handles .Focus by selecting all text in the textbox
+                caretIndex = textBox.CaretIndex;
+            }
             Description.UnsavedChanges = !saved;
             Text = Description.DisplayNameWithStatus;
-            caller?.Focus();
+            if (caller is not null && caller is TextBox box)
+            {
+                box.Focus();
+                box.CaretIndex = caretIndex;
+            }
+            else
+            {
+                caller?.Focus();
+            }
         }
     }
 }
--- a/src/SerialLoops/Editors/Editor.cs
+++ b/src/SerialLoops/Editors/Editor.cs
@@ -35,6 +35,11 @@ namespace SerialLoops.Editors
 
         public void UpdateTabTitle(bool saved, Control caller = null)
         {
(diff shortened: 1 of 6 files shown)


src/SerialLoops/Editors/Editor.cs
```diff
@@ -32,11 +32,26 @@ void InitializeComponent()
...
             Text = Description.DisplayNameWithStatus;
+            if (caller is not null && caller is TextBox box)
+            {
+                box.Focus();
+                box.CaretIndex = caretIndex;
+            }
+            else
+            {
+                caller?.Focus();
+            }
```
WiIIiam278, line 54: Unfortunately this seems to have a slight performance hit that can be tested by holding down a key on this version vs the previous version; text takes slightly longer to enter. Given how often this is being called, perhaps a check could be added if the title has already been updated to not carry out this refocusing logic each time if the title doesn't need to be updated?

WiIIiam278 (approved):
I can't test all the x-plat fixes, but this does seem to work :)

src/SerialLoops/Editors/Editor.cs
jonko0493: that's extremely smart!

PR #81, opened 2023-04-02
Fixup macOS patch creation

### LINKED ISSUE #80: Patch creation on Mac results in crash
On Windows and Linux, xdelta patches can be created without issue.

On macOS, I seem to instead get the following exception on `encoder.Encode()` (line 14 in Patch.cs):

```
An unhandled exception of type 'System.AccessViolationException' occurred in VCDiff: 'Attempted to read or write protected memory. This is often an indication that other memory is corrupt.'
```

idk seems bad man

jonko0493:
Fixes #80. Turns out it was a pretty simple fix lol.


WiIIiam278 (approved):
tested on my mac, works

PR #86, opened 2023-04-03
Add crash handler on Windows, enable crash logging on Mac

### LINKED ISSUE #79: Add crash handler
We want to be able to get logs from crashes, obviously for when people start using the program in the wild, but also because we usually run it sans-debugger on Mac.

A crash handler could probably be very simple -- something like a try/catch around the main program loop would work, as long as it's capable of logging and (potentially) attempting to save the project before closing.

jonko0493:
Linux: 🤷‍♀️ 

This adds a simple crash handler on Windows that can catch exceptions prior to a crash and log them to the log file. We add a special `LogCrash` function to the `LoopyLogger` so that we don't pop a message box when logging that particular error.

WPF is able to catch these exceptions because they are its native exceptions; however, we're not so lucky on Mac and Linux. The end result here is that we need to be better at exception handling generally. In the meantime though, I've enabled crash logging on Mac which will give stack traces in the standard macOS "Report this error" dialog, so we can tell people to do that to get useful logs on that platform at least.

Fixes #79.

### DIFF
--- a/src/SerialLoops/Utility/LoopyLogger.cs
+++ b/src/SerialLoops/Utility/LoopyLogger.cs
@@ -52,5 +52,14 @@ namespace SerialLoops.Utility
                 _writer.Flush();
             }
         }
+
+        public void LogCrash(Exception ex)
+        {
+            if (_writer is not null)
+            {
+                _writer.WriteLine($"{DateTimeOffset.Now} - ERROR: {ex.Message}\n\n{ex.StackTrace}");
+                _writer.Flush();
+            }
+        }
     }
 }
--- a/src/SerialLoops/Utility/LoopyLogger.cs
+++ b/src/SerialLoops/Utility/LoopyLogger.cs
@@ -57,7 +57,7 @@ namespace SerialLoops.Utility
         {
             if (_writer is not null)
             {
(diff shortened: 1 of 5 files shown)


src/SerialLoops/Utility/LoopyLogger.cs
```diff
@@ -52,5 +52,14 @@ public void LogWarning(string message, bool lookForErrors = false)
                 _writer.Flush();
             }
         }
+
+        public void LogCrash(Exception ex)
+        {
+            if (_writer is not null)
+            {
+                _writer.WriteLine($"{DateTimeOffset.Now} - ERROR: {ex.Message}\n\n{ex.StackTrace}");
```
WiIIiam278, line 60: We might want to verbosely state "SERIAL LOOPS HAS CRASHED" or something to make it really obvious in the log file

WiIIiam278 (approved):
I haven't tested this yet but the implementation looks good. Haven't had a chance to see why checks are failing

src/SerialLoops/Utility/LoopyLogger.cs
jonko0493: Good idea; implemented.

PR #92, opened 2023-04-03
Fixup workspace tabs

### LINKED ISSUE #85: Tabs are loaded off screen when reloading project workspace
Works fine on Linux, but Mac and Windows definitely shouldn't have this. We had a similar bug before just with the tabs opening, but I don't remember exactly what fixed that. Do you, @WiIIiam278?

jonko0493:
Fixes #85. Also, while we're at it, removes the badge labels because they do something different than I thought they did lol.

Also, while we're here, ensure that the "close on saving" function actually commits the changes by building iteratively.

### DIFF
--- a/src/SerialLoops/MainForm.eto.cs
+++ b/src/SerialLoops/MainForm.eto.cs
@@ -203,6 +203,10 @@ namespace SerialLoops
                         }
                         tracker.Finished++;
                     }
+                    if (EditorTabs.Tabs.Pages.Count > 0)
+                    {
+                        EditorTabs.Tabs.SelectedIndex = 0;
+                    }
                 }
             }
             catch (Exception e)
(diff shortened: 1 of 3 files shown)


src/SerialLoops/MainForm.eto.cs
```diff
@@ -203,6 +203,10 @@ private void LoadCachedData(Project project, IProgressTracker tracker)
                         }
                         tracker.Finished++;
                     }
+                    if (EditorTabs.Tabs.Pages.Count > 0)
+                    {
+                        EditorTabs.Tabs.SelectedIndex = 0;
```
WiIIiam278, line 208: While this certainly does fix the issue, the way the workspace re-opener is setup is such that it'll always reopen the last tab you had open -- and this breaks that functionality somewhat. I wonder if perhaps we could set the index to the last one _again_ after this to fix it fully? Not sure

jonko0493: No, haha, I think we have to settle for having it open the first tab. Another possible solution, however, could be to open the tab that was opened last as the first tab?

WiIIiam278 (approved):
approving to not block! following on from discord convo, we realised the tab index you had open isnt actually being saved anyway so its okay to just open the first in the list to solve this issue

PR #112, opened 2023-04-04
Fix home screen not opening when project fails to load due to corruption

### LINKED ISSUE #58: Corrupted projects don't load splash screen after failing to load
If a project is corrupted (i.e. one of the archives has an invalid file in it that breaks on loading), it will fail to load. It should then show us the splash screen. Currently, we just end up with a blank screen.

WiIIiam278:
Fixes #58


WiIIiam278:
I have no idea why this made the windows test explode lol.. gonna try re-running. if not, will need to test again on my windows machine more thoroughly when I get home

jonko0493:
The windows test is just flaky. You have to re-run it occasionally. Maybe one day I'll figure out why lol

jonko0493 (approved):

PR #127, opened 2023-04-06
Create update-checker app (LoopyUpdater), add prefs for auto updating

### LINKED ISSUE #120: Add built-in updater
The design concept for this:

On launch, Serial Loops checks GitHub (or if the API is finnicky with this, perhaps we can provide a versioning JSON on our site in the same style as the RSS feed). If a new release is detected, SL prompts the user to see if they want to download it -- if they do, then a simple CLI app (or perhaps another Eto app) is launched and SL shuts itself down. After waiting a bit, the updater program downloads the SL artifacts, unpacks them, and then moves them into the SL directory.

WiIIiam278:
Closes #120 (at least for MVP, should prob split that up into more issues)

Still need to figure out invoking on shutdown & this doesn't yet actually do the update pulling (though doing so shouldn't be too tricky). In the future I'd definitely like to convert this to a super mini eto app.


LoopyUpdater/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        // todo
+        private void UpdateMacApp(string latestVersion)
+        {
+            DisplayUpdateMessage($"macOS ({RuntimeInformation.OSArchitecture.ToString()})", latestVersion);
+        }
+
+        private void DisplayUpdateMessage(string platform, string latestVersion)
+        {
+            Console.WriteLine("\n******************************\n" +
```
jonko0493, line 83: As I mentioned in chat, I'd probably do this part in Serial Loops itself. I think the updater should only handle actually updating the binaries themselves.

src/SerialLoops/MainForm.eto.cs
```diff
@@ -560,5 +559,15 @@ public void CloseProject_Executed(object sender, EventArgs e)
...
             }
         }
+
+        public void AppClosed_Executed(object sender, EventArgs e)
+        {
+            if (!CurrentConfig.AutomaticUpdates) return;
+            
+            string currentVersion = Assembly.GetExecutingAssembly().GetName().Version?.ToString() ?? "0.1";
+            string executableDirectory = Path.GetFullPath(Assembly.GetExecutingAssembly().Location);
+            //todo execute LoopyUpdater
```
jonko0493, line 569: Should be something like `Process.Start("LoopyUpdater.exe")` but it will be interesting to see how that works when the parent process is getting closed. We'll have to figure something out there. Perhaps we launch it via cmd/bash (platform dependent) or something like that

jonko0493 (commented):
Good start, but still needs some work before merging, I think.

jonko0493:
To be more specific: I think if you take the update checking logic and put that directly into SerialLoops on startup and then scaffold the basics for the updater CLI, I think this would be good to merge!

LoopyUpdater/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+        public static void Main(string[] args)
+        {
+            if (args.Length == 0)
+            {
+                Console.WriteLine("Invalid syntax.\n- Usage: LoopyUpdater <version>");
+                return;
+            }
+
+            var executableDirectory = args.Length >= 2
+                ? string.Join(" ", args[1..args.Length])
```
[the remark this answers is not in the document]
WiIIiam278: Oh, huh, I figured it split on spaces and didn't realise that args[1] would work to return the second quoted arg :)

src/SerialLoops.Lib/Config.cs
```diff
@@ -102,7 +104,9 @@ private static Config GetDefault(ILogger log)
                 EmulatorPath = emulatorPath,
                 AutoReopenLastProject = true,
                 RememberProjectWorkspace = true,
-                RemoveMissingProjects = false
+                RemoveMissingProjects = false,
+                CheckForUpdates = true,
+                PreReleaseChannel = true
```
jonko0493, line 109: Set to false by default

jonko0493: also, the nittiest of nits, but should be
```suggestion
                PreReleaseChannel = false,
```

the comma makes it so next time you update the list, that line doesn't get included in the diff

src/SerialLoops/MainForm.eto.cs
```diff
@@ -631,5 +644,21 @@ public void CloseProject_Executed(object sender, EventArgs e)
                 ProjectsCache.Save(Log);
             }
         }
+
+        public void AppClosed_Executed(object sender, EventArgs e)
+        {
+            if (ShutdownUpdateUrl is not null)
+            {
+                string updaterExecutable = "LoopyUpdater.exe";
```
jonko0493, line 652: maybe make this a const?

src/SerialLoops/Utility/UpdateChecker.cs
```diff
@@ -0,0 +1,82 @@
...
+                new ProductInfoHeaderValue(USER_AGENT, currentVersion)
+            );
+
+            try
+            {
+                var response = await client.GetAsync(GET_RELEASES);
+                var content = await response.Content.ReadAsStringAsync();
+                var json = JsonNode.Parse(content).AsArray();
+
+                foreach (JsonNode release in json)
```
jonko0493, line 65: ngl, genuinely impressed you did this with System.Text.Json

jonko0493 (approved):
Couple more comments, but approving to unblock.

Before merging, please comment out all the logic for actually calling the updater and replace it with a button that links to the release page for now. I think that's a good MVP!

src/SerialLoops/Utility/UpdateChecker.cs
WiIIiam278: so am I lol

WiIIiam278:
Yeah cool -- the scaffold for the updater app is sorta there at least with this and best not let the scope creep up too much

PR #128, opened 2023-04-06
Add official build pre-releases

### LINKED ISSUE #67: Serial Loops icon doesn't display in system tray on Ubuntu
image

Probably just some configuration issue we need to sort out.

### LINKED ISSUE #126: Add official build pre-releases
We should update a pre-release channel with builds on each merge to main.

jonko0493:
Closes #126. This completely redoes the official build so that it produces a GitHub release on every commit to main. It produces a deb package for Debian-based Linux distros, a tar.gz with published binaries for other Linux distros, two different dmgs for the x64 and ARM flavors of macOS, and a zip file for Windows. In the future, we might consider building an MSI in CI or installer for Windows and publishing that as well.

Additionally, this fixes #67 as the dpkg adds a `.desktop` file to `/usr/share/applications` that makes the launch tray contain the SerialLoops icon.


azure-pipelines-official.yml
```diff
@@ -4,37 +4,175 @@ trigger:
...
+      inputs:
+        gitHubConnection: 'GitHub Connection (Jonko)'
+        tagSource: userSpecifiedTag
+        tag: '$(Version)'
+        title: 'Serial Loops v$(Version)'
+        releaseNotesSource: inline
+        releaseNotesInline: |
+          Latest build from main. Note that this is likely to have bugs and we recommend you use a regular release instead!
+
+          On macOS, after dragging the app from the dmg into Applications, please run `xattr -cr /Applications/SerialLoops.Mac.app` from the Terminal in order to be able to run the app.
```
WiIIiam278, line 175: Is this to get around macOS' GateKeeper? iirc can't you also just go to Privacy Settings and "Open Anyway"

Also -- maybe we should remind peeps of the DKP requirement here lol

WiIIiam278 (approved):
nice :)

azure-pipelines-official.yml
jonko0493: for the first -- I think I did that already on my mac but still ran into a problem where it told me the application was "damaged" and advised me to trash it. Running this command fixed it.

for the second -- yes, agreed, very good idea. I'll update accordingly

WiIIiam278: Out of interest would running that on the DMGs on the mac runner before we upload fix it?

jonko0493: That's a good question. I don't think so, because I'm pretty sure it's a machine-specific attribute, but perhaps it's worth testing?

PR #148, opened 2023-04-11
Add script templates feature, add one basic script template

### LINKED ISSUE #102: Add script file templates
Started with the idea for chess script files, but it actually would make sense to make this broader.

Chess files are a little bit difficult to create with the editor right now. We should have an option to (when you have a blank script file) apply a "chess script template" to make things easier.

We could then also add additional templates for regular script files, topic files, etc. Stuff that puts the basic scaffolding in place to make the boilerplate stuff easier/go away.

jonko0493:
Closes #102, but should probably open a follow-up issue for a future milestone to add more templates.

Adds "script template" functionality, which basically is the ability to quickly add a set of predefined commands to your script. The basic one I've created is for adding the most common boilerplate commands to `SCRIPT00`. In the future, the functionality can and should be expanded.

@WiIIiam278 there are a lot of UI improvements here that could be made -- most notably, we need an icon for apply templates. Please feel free to commit any changes to UI directly to this branch.


src/SerialLoops.Lib/Script/ScriptTemplates.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                Name = name;
+                Description = description;
+                Template = template;
+            }
+        }
+
+        public static readonly ObservableCollection<TemplateOption> AvailableTemplates = new()
+        {
+            new("Standard Story Script Intro", "A standard intro for a mainline script (i.e. EVX_XXX)", (Project p) => StandardStoryScriptIntro(p)),
```
WiIIiam278, line 29: perhaps "story script" instead of "mainline script"

src/SerialLoops/Editors/ScriptEditor.cs
```diff
@@ -77,13 +78,29 @@ private Container GetCommandsContainer()
...
             {
-                var barCommand = EditorCommands.Find(barCommand => barCommand.ToolBarText.ToLower().Equals(command));
+                var barCommand = EditorCommands.Find(bc => bc.ToolBarText.ToLower().Equals(command));
                 if (barCommand is not null)
                 {
                     treeGridView.Control.MapPlatformCommand(command, barCommand);
                 }
             }
 
+            Command applyTemplate = new() { MenuText = "Apply Template", ToolBarText = "Apply Template", Image = ControlGenerator.GetIcon("AppIcon", _log) };
```
WiIIiam278, line 88: Perhaps "Templates" instead of "Apply Template"? Otherwise we'll have to rejig the taskbar because this runs onto a new line. Still makes sense that way imo since it opens the Templates menu. Maybe also a button for this next to the 🚀 on the script editor itself?
tbh should add the rocket up here too as there's a bit of redundancy with needing to define two sets of buttons/commands. might be a "william's next PR" thing lol

WiIIiam278 (approved):
nice work :) 

Only thing I would say is that if you apply a template on an existing file it can be a bit confusing -- maybe some templates could be marked as "starter" templates and give a warning that adding them to a file may have unintended consequences?

also I'd say when we add new script creating we should display the template panel as the first thing you see (& can show starter templates on there)

src/SerialLoops.Lib/Script/ScriptTemplate.cs
```diff
@@ -0,0 +1,66 @@
+﻿using HaruhiChokuretsuLib.Archive.Event;
+using SerialLoops.Lib.Items;
+
+namespace SerialLoops.Lib.Script
+{
+
+    public class ScriptTemplate
```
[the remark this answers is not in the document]
jonko0493: This is a good idea. Do you mind filing an issue for it?

PR #155, opened 2023-04-12
Fix freeze on mac when nesting progress dialogs

### LINKED ISSUE #154: Softlock when attempting to build unbuilt items when opening project on macOS
Attempting to build unbuilt items when opening a project on macOS will cause the second progress dialog to open and get stuck on "Loading Project Settings" causing the app to freeze. requires a force-quit through activity monitor to close

jonko0493:
Fixes #154.

The diff is a little confusing, but what's happening here is I moved the progress dialogs out of the _onComplete action for the first progress dialog, which means there's no conflict in threads now.


WiIIiam278 (approved):
nice :+1:

PR #160, opened 2023-04-13
Add a few small bugfixes

jonko0493:
Honestly don't even remember what all these do but they're important, trust me.

Edit:
* First prevents a crash due to nullref when doing something with map characters, idk
* Second just changes the name of a parameter to make more sense
* Third changes Haruhi Meter to actually work (first parameter of `HARUHI_METER` command is unused so we need to adjust for that)


WiIIiam278 (approved):
i'm not quite sure what these do either lol but cool

PR #164, opened 2023-04-16
Try fixing versioning issues, maybe

### LINKED ISSUE #163: Update checker thinks there's always a new version of SL
Internal mismatch between assembly version and what the update checker is expecting.

jonko0493:
Fixes #163 (maybe).

Introduces a build.props to ensure all our assemblies have a consistent version and that we can set it from within CI. Also removes PDBs from release builds, which is good.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+
+  <PropertyGroup Condition="'$(SLVersion)' == ''">
+    <SLVersion>0.2.9999</SLVersion>
+  </PropertyGroup>
+
+</Project>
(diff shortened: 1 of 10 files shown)


src/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+
+  <PropertyGroup Condition="'$(SLVersion)' == ''">
+    <SLVersion>0.2.9999</SLVersion>
```
WiIIiam278, line 5: Kind of cursed but.. yeah, okay!!!

WiIIiam278 (approved):
I think this works?! I guess we won't truly know until it stops working when we bump the version number again lol

PR #171, opened 2023-04-18
Change build number to fixup versioning

jonko0493:
Fixes an error in the pre-release builds with the semantic versioning being too high a number. This breaks out the build name a bit more which will make the version happier.


jonko0493:
@WiIIiam278 u know it's just the usual transient failure lol; requeued

WiIIiam278 (approved):

PR #174, opened 2023-04-23
Fix SET_PLACE crash, fix BG replacement on low color images, fix versioning scheme for pre-releases

### LINKED ISSUE #172: Crash when setting SET_PLACE to NONE
Tested on Win...crashes when you pick "NONE" as the place name to display 😢

### LINKED ISSUE #173: "Sequence contains no elements" error when replacing a background with a grayscale image
This error appears whenever I attempt to replace a background with a new one with a lower resolution:

image

We should probably show a more sensible error if you try and do this.

jonko0493:
Fixes #172. Fixes #173. 

Also updates versioning scheme for pre-releases to make more sense, and updates the about dialog to use the friendly informational version rather than the strict assembly version.

Also fixes a small bug with rendering BGs where I was declaring height to be 194px rather than 192px lol.

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -18,14 +18,19 @@ parameters:
   type: string
   default: 'Latest nightly build. **Note that this is likely to have bugs and we recommend you use a regular release instead!**'
 
-name: $(Year:yy).$(Date:MMdd).$(Rev:r)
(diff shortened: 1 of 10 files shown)


WiIIiam278 (approved):
Tested -- can replace B&W images just fine now :) Version seems to display just fine in the about menu, too.

PR #177, opened 2023-04-27
Use improved color reduction algorithm

### LINKED ISSUE #111: Improve color reduction algorithm
There are a variety of options for color reduction. The one we went with is just fine for the UI changes we were making, but imo flattens colors a bit too much (especially in gradients) for CGs and such. We should look into a replacement that is superior in that regard.

As an example, above is original and below is replaced:

EV550

EV550_replaced

Lib changes: the palette stuff is handled at the lib level, so that's where the change would need to happen.

### LINKED ISSUE #175: Crash when setting KBG_DISP to NONE

jonko0493:
Closes #111. I'd put before/after photos, but you literally can't tell the difference anymore. 😎

Separately, fixes #175.

### DIFF
--- a/src/SerialLoops.Lib/Items/BackgroundItem.cs
+++ b/src/SerialLoops.Lib/Items/BackgroundItem.cs
@@ -96,20 +96,21 @@ namespace SerialLoops.Lib.Items
             }
         }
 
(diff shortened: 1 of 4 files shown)


WiIIiam278 (approved):
Oh I forgot to review this

PR #178, opened 2023-05-10
Add ability to rename items

### LINKED ISSUE #88: Add ability to rename items
I'd recommend "faking" this at a project level and basically creating a map file that we load in that maps the "project name" of an item to its internal name. Will make things much easier for projects where items are being replaced or just for folks who want to be able to keep track of things.

Potentially, with this model, we could also provide a set of default renames.

jonko0493:
Closes #88.

When selecting an item, you can right click -> Rename or hit ctrl-R to rename it now!

Additionally, I've added default names to several of the items that will be loaded in on project creation.


src/SerialLoops/Dialogs/ItemRenameDialog.cs
```diff
@@ -0,0 +1,107 @@
...
+            _project = project;
+            _log = log;
+            InitializeComponent();
+        }
+
+        private void InitializeComponent()
+        {
+            MinimumSize = new(256, 100);
+
+            TextBox nameBox = new();
```
WiIIiam278, line 29: How about putting the current name of the item as the default in here and focusing the box? Fairly often I'll probably just want to adjust the name by a letter or two I imagine so this'd be nice.

src/SerialLoops/Controls/ItemExplorerPanel.cs
```diff
@@ -23,6 +22,13 @@ public ItemExplorerPanel(Project project, EditorTabsPanel tabs, ILogger log) : b
             {
                 ((TreeGridView)Viewer.Control).ContextMenu = Viewer.SelectedItem.Text.GetContextMenu(_project, this, _tabs, _log);
             }
+            KeyUp += (sender, args) =>
+            {
+                if (args.Control && args.KeyData.HasFlag(Keys.R))
```
WiIIiam278, line 27: This is totally fine and works but I think this should be added as a command to the menu bar somewhere (perhaps under "File"). I think Eto wants us to use the built in shortcuts API for this sort of stuff -- For instance, this doesn't feel nice on macOS where the CMD key is the general modifier. It'd also sort of be nice to have it in the menu bar anyway

jonko0493: Ah, so it doesn't automatically fixup the modifier across platforms... yeah I will do the shortcuts thing since I think that is definitely superior!!

WiIIiam278: Maybe we should add a way of customising the shortcuts in settings...maybe. I'm more of a fan of f2 for renaming

jonko0493: That sounds like something in your wheelhouse. :P

File an issue and pick a milestone and we'll do it 😎

src/SerialLoops/Dialogs/ItemRenameDialog.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                string name = $"{prefix}{nameBox.Text.Trim()}";
+                if (_project.Items.All(i => !i.DisplayName.Equals(name, StringComparison.OrdinalIgnoreCase)))
+                {
+                    _item.Rename(name);
+                    Close();
+                }
+                else
+                {
+                    _log.LogError($"Name '{name}' is already in use!");
```
WiIIiam278, line 71: Interesting thing: Pressing ENTER to dismiss this error prompt autofocuses the text bar and immediately re-enters the rename, causing an infinite loop of doom. Manually clicking Okay to dismiss works, though.

src/SerialLoops/Utility/Shared.cs
```diff
@@ -0,0 +1,37 @@
...
+using SerialLoops.Lib.Items;
+using System.Linq;
+
+namespace SerialLoops.Utility
+{
+    public static class Shared
+    {
+        public static void RenameItem(Project project, ItemExplorerPanel explorer, EditorTabsPanel tabs, ILogger log, bool overrideRename = false)
+        {
+            ItemDescription item = project.FindItem(explorer.Viewer.SelectedItem?.Text);
```
WiIIiam278, line 15: This causes a crash when no item is selected in the sidebar (possible on macOS 😎) since it's trying to FindItem with `null` as the name string

WiIIiam278 (approved):
works great apart from those two (easy) fixes :D 
I'm unable to find any other bugs -- nice!

src/SerialLoops/Dialogs/ItemRenameDialog.cs
jonko0493: Yes I noticed this as well on Windows. It's weird, but I'm not sure what the fix is. Because manually clicking works I'm going to leave it for now.

src/SerialLoops/Utility/Shared.cs
jonko0493: Fixed

PR #181, opened 2023-05-13
Replace "dialogue config" with "character" abstraction that can be used instead of Speakers and add nameplate editor

### LINKED ISSUE #91: Replace Dialogue Config items with "Character" items and allow editing nameplates
Dialogue config items essentially just represent the properties of a particular Speaker (i.e., their nameplate location, voice font, and text timer). We should start representing them as characters and allow editing nameplates.

1. Create a mapping between the internal speaker enum and the speaker display name that gets stored at the project level
2. Implement a nameplate editor which allows for directly editing the nameplate file using text input only. It should be able to intelligently squish the text for longer names. The nameplate editor is also where you should be able to define character aliases.

jonko0493:
Closes #91.


src/SerialLoops.Lib/DefaultCharacters.en.json
```diff
@@ -0,0 +1,177 @@
...
+    "PlateColor": "FF707070",
+    "OutlineColor": "FF353535",
+    "HasOutline": true
+  },
+  "23": {
+    "Name": "Information",
+    "NameColor": "00000000",
+    "PlateColor": "00000000",
+    "OutlineColor": "FF353535",
+    "HasOutline": true
```
WiIIiam278, line 161: What we should probably do for Monologue and Information is set the name color to white with a sensible plate color default so that when users edit it they don't get a glitchy black box thing

WiIIiam278 (approved):
Amazing :) Two things as well:

Only other thing is that it's a bit weird that we're using both an underscore and a space in the names, would be nice to have them consistent (either always w/ or w/o an underscore)
image

src/SerialLoops.Lib/DefaultCharacters.en.json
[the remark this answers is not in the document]
jonko0493: Yep yep, agreed!

src/SerialLoops.Lib/DefaultCharacters.en.json
jonko0493: I have an even better idea: prevent these two from being edited at all. It won't matter once we're able to add new speakers, and these two are special anyway.

jonko0493:
As for the spaces & underscores: honestly, I don't want to write something that translates between the two. I think it's hacky and ultimately unnecessary. I'm okay with the stylistic inconsistency, personally.

PR #184, opened 2023-05-14
Add editing for place names

### LINKED ISSUE #95: Make place items editable
Will likely use the same tech as #91, we should be able to type in the text for place names.

Goes hand-in-hand with #88.

jonko0493:
Closes #95.

image


src/SerialLoops/Editors/PlaceEditor.cs
```diff
@@ -17,13 +22,46 @@ public PlaceEditor(PlaceItem placeItem, Project project, ILogger log) : base(pla
         public override Container GetEditorPanel()
         {
             _place = (PlaceItem)Description;
+            if (string.IsNullOrEmpty(_place.PlaceName))
+            {
+                _place.PlaceName = _place.DisplayName[4..];
+            }
+
+            TextBox placeTextBox = new() { Text = _place.DisplayName[4..] };
```
WiIIiam278, line 30: It seems like DisplayName added to this box is always just the item name -- possibly because DisplayName isn't being set anywhere? Bare in mind I tested using an existing project.
image

src/SerialLoops.Lib/Items/PlaceItem.cs
```diff
@@ -51,16 +53,101 @@ public SKBitmap GetPreview(Project project)
...
+        {
+            return Unscramble(PlaceGraphic.GetImage(transparentIndex: 0));
+        }
+
+        public SKBitmap GetNewPlaceGraphic(SKTypeface msGothicHaruhi)
+        {
+            string spaceAdjustedText = PlaceName.Replace(" ", "   ");
+            SKBitmap newPlaceBitmap = new(PlaceGraphic.Width, PlaceGraphic.Height);
+            SKCanvas canvas = new(newPlaceBitmap);
+            SKColor bgColor = new SKColor(0, 249, 0);
```
WiIIiam278, line 69: It might be nice to add a bool ("preview") flag to this method -- if set, instead of painting the BG green it'll just be transparent. That way, this can be passed to render a preview without the green BG, while the method can be passed without the preview bool to get the real image.

WiIIiam278 (approved):
Works great!

src/SerialLoops/Editors/PlaceEditor.cs
jonko0493: Testing using an existing project is the problem here, I believe -- the place names are set in the default names spot.

src/SerialLoops.Lib/Items/PlaceItem.cs
jonko0493: I kind of disagree here tbh, bc `GetNewPlaceGraphic` shouldn't be used for previews, imo. There's a separate `GetPreview` method that does that; `GetNewPlaceGraphic` should only be used for replacing the place graphic.

jonko0493:
Going to merge for now, but willing to change my mind on some feedback with further discussion!

PR #187, opened 2023-06-01
Add voiced line subtitle editing

### LINKED ISSUE #105: Add editing for voice line subtitles
We can detect the existence of VOICEMAP.S (the custom file we made for this) and then add subtitle editing to the voice line items.

Choku lib change: subtitle file needs to be redesigned to allow easy adding/removing subtitles.

jonko0493:
Closes #105.

Adds subtitle editing for voiced lines. There is a limitation here as we had to adjust how the subtitles file worked in order to add the capability of adding new subtitles; thus, this functionality doesn't work with rom versions 0.4 and below, only 0.5 (0.6 public release) and above.

image


WiIIiam278 (approved):
Builds and runs, will need to test functionality when the runner's built todays build (which I assume adopts the new sub format?), but unblocking as this looks fine :)

PR #189, opened 2023-06-02
Add preview for voiced line subtitle editor

### LINKED ISSUE #188: Add subtitle preview
We should expand on #187 and add a preview for subtitles to make it easier to see where they'll show up on the screen. All the information for drawing them is contained in the subtitle structs, so it shouldn't be difficult to implement this.

* The struct contains a top/bottom screen enum which can be used to determine which screen to draw to
* The struct contains X/Y coordinates for drawing position on the screen
* The actual drawing routine is as follows:
  * On the top screen, the subtitle is simply drawn in white text at (X,Y)
  * On the bottom screen, the subtitle is drawn in white at (X,Y) and then drawn in black at (X+1,Y+1) and (X,Y+1) behind the white text
* Ensure you use `.Subtitle` so that the formatting (`#P07`) is stripped

WiIIiam278:
Closes #188

Adds a text preview to the subtitle editor when modifying subtitles and improves the UI

image


src/SerialLoops/Editors/VoicedLineEditor.cs
```diff
@@ -153,30 +153,92 @@ public override Container GetEditorPanel()
...
+            SKBitmap previewBitmap = new(256, 384);
+            SKCanvas canvas = new(previewBitmap);
+            canvas.DrawColor(SKColors.DarkGray);
+            canvas.DrawLine(new SKPoint { X = 0, Y = 192 }, new SKPoint { X = 256, Y = 192 }, DialogueScriptParameter.Paint00);
+
+            var voiceMapStruct = _project.VoiceMap.VoiceMapStructs.FirstOrDefault(v => v.VoiceFileName == Path.GetFileNameWithoutExtension(_vce.VoiceFile));
+            bool bottomScreen = voiceMapStruct.TargetScreen == VoiceMapFile.VoiceMapStruct.Screen.BOTTOM;
+            if (bottomScreen)
+            {
+                ScriptEditor.DrawText(
```
jonko0493, line 218: ooh, actually this needs to be drawn twice -- there's also an (X+1, Y) shadow that's drawn.

jonko0493 (changes requested):

jonko0493 (approved):
Very clever for loop, love it.

PR #190, opened 2023-06-03
Add topic editing and the fruits of RE labor

### LINKED ISSUE #99: Add topic editing
We should be able to edit the topics.

RE work: Determine how the main topic weird dance of associations works. Determine what at least some of the unknowns do.

Choku lib work: reflect RE work.

jonko0493:
Closes #99.

The headliner of this PR is the ability to edit topics. Specifically, topic names, linked scripts, and times can now be edited.

Notably, this does not include the ability to change topic _types_. This is because the type of the topic is inherently tied to its ID, which is actually an internal flag. Additionally, the work here helped figure out how to associate the duplicated main topics (ID + 40), so those are combined into one item, bringing the total topic count to 511 (510 topics present in-game + 1 dummy topic).

Doing the RE work here pulled back some other layers of what other things do, and unveiled that several other unknown values were also flags. Additionally, the nature of the `GLOBAL` command was revealed to actually be something that modifies "friendship levels" and thus has been renamed `MODIFY_FRIENDSHIP`.


src/SerialLoops.Lib/Script/Parameters/FriendshipLevelScriptParameter.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public class FriendshipLevelScriptParameter : ScriptParameter
+    {
+        public enum FriendshipCharacter : short
+        {
+            Haruhi = 2,
+            Mikuru = 3,
+            Nagato = 4,
+            Koizumi = 5,
+            Tsuruya = 7,
```
WiIIiam278, line 13: Are there characters missing here we don't know the IDs of yet? Interesting that you can befriend Tsuruya!

src/SerialLoops/Editors/TopicEditor.cs
```diff
@@ -105,109 +182,104 @@ public override Container GetEditorPanel()
...
+            //        ControlGenerator.GetControlWithLabel("Unknown 03", new TextBox { Text = _topic.Topic.UnknownShort03.ToString() }),
+            //        ControlGenerator.GetControlWithLabel("Unknown 04", new TextBox { Text = _topic.Topic.UnknownShort04.ToString() }),
+            //    }
+            //};
 
-            return new TableLayout(ControlGenerator.GetControlWithLabelTable("ID", idLabel),
+            return new TableLayout(idLayout,
                 ControlGenerator.GetControlWithLabel("Title", titleTextBox),
+                ControlGenerator.GetControlWithLabel("Type", _topic.Topic.CardType.ToString()),
                 ControlGenerator.GetControlWithLabel("Associated Script", linkedScriptDropDown),
```
WiIIiam278, line 248: I think a link to the selected script next to the dropdown would be very handy here

WiIIiam278 (approved):
Just tested -- seems to work great! Amazing stuff :)

src/SerialLoops.Lib/Script/Parameters/FriendshipLevelScriptParameter.cs
jonko0493: there is a 6th slot for friendship levels, but it can't be referenced as a variable so I ignored it. I'll add a comment here.

WiIIiam278: pretty random that Tsuruya is just also sort of befriendable. love it, though!

Issue #106, opened 2023-04-03
Add "advanced search"


jonko0493:
Add the ability to specify what you're searching for when searching through files.

Examples:
* Search scripts by conditional
* Search scripts by flags
* Search scripts by various command parameters
* Search backgrounds by type

PR #194, opened 2023-06-06
Use screen selector in script editor, add quick search to main item explorer, minor tweaks

### LINKED ISSUE #162: Fixup chibi direction facing buttons on Linux
They're too big to click them all.

image

WiIIiam278:
Does a bunch of things I've been meaning to do:
* Adds the new button-style script selector to the script editor, with new "both" checkbox to select both screens
* Adds a quick search bar to the main project view, which filters only by item titles.
  * I'd like to make the SearchDialog much more powerful with more advanced filtering quite soon -- sort of as an "advanced search." I think having a quick search filter on the left bar is helpful and easy to use. 
  * I've also made it so if you enter text in there and hit CTRL+F to bring up the dialog, your text in that field will carry over if you decide you need a more advanced search.
* Changes search boxes to use the actual Eto.Forms SearchBox instead of TextBox (giving us nice platform-specific styles!)

image


WiIIiam278:
(it looks like rider got a bit too excited and autoformatted the mainform class. I actually think it seems to have done a decent job at catching a few weird and inconsistent bits of formatting, but if there's anything odd let me know and i'll fix it up)

jonko0493:
Search bar in the item explorer rules. Implementing advanced search is #106, so very down if you take that on next! :)

This is just excellent work all around. I'll get to testing it (especially on Linux) in the next few hours.

jonko0493:
As for the autoformatting of the main class: it's fine, I'll run it through VS and reformat it once this merges :P

src/SerialLoops/Controls/ChibiDirectionSelector.cs
```diff
@@ -62,7 +62,9 @@ private void InitializeComponent()
                 Image = ControlGenerator.GetIcon($"Chibi_{direction}", _log),
                 Enabled = AvailableDirections.Contains(direction),
                 Tag = direction,
-                Size = new(35, 35)
+                Size = new(35, 35),
+                MinimumSize = new(35, 35),
```
jonko0493, line 66: Unfortunately, for whatever reason this still doesn't fix this on Linux.

image

Personally, I'd recommend VirtualBox + lubuntu for testing this

jonko0493 (commented):
This might be difficult, so if it's too hard just tell me no lol. But currently, when you search, it will collapse all the categories. Would it be possible to preserve expansion or at least expand all categories when searching?

jonko0493 (approved):
Looks good; issue #198 filed for feedback regarding remembering state change for categories on search.

PR #195, opened 2023-06-11
Add system texture items

### LINKED ISSUE #104: Add system texture items
There are a huge variety of system textures in SYSTEX.S -- we should make sure at the ones that make sense to replace get added as system texture items. This will require more reverse-engineering work to determine what exactly the various system texture components are/do.

Part of this could be hardcoded -- for example, adding the ability to edit logo splash screens as their own unit would be interesting.

jonko0493:
Closes #104.

This adds system texture items to the editor and allows exporting and replacing them just like backgrounds. The system texture items added so far are the logos/credits screen, the extras items, and all textures prefixed with `SYS`. System textures are very weird and technical, so I don't think at this point it makes sense to support actually editing the SYSTEX.S file. Additionally, we give them all default names here.


WiIIiam278 (approved):
This is excellent, works flawlessly! Note I tested with a 0.5 ROM. A few tiny nitpicks nonetheless:

* It'd be nice if the splash images were grouped neatly, sharing a "SPLASH_" prefix so they're adjacent to one another
* SYSTEX_DIALOGUE_H seems to be duplicated 
image

jonko0493:
Fixed the first thing. Looking into SYSTEX_DIALOGUE_H being duped.

jonko0493:
Most recent change fixes the location of the distinct

PR #196, opened 2023-06-11
Add UI text editing

### LINKED ISSUE #159: Add UI text editing
Possibly something someone might find interesting? Not sure, but we can look into adding it at some point.

jonko0493:
Adds UI text editing capabilities. The technique here is to just have them in a dialog that we pull up and allow editing them directly. Super simple but works well!

Closes #159.

image


src/SerialLoops/Dialogs/EditUiTextDialog.cs
```diff
@@ -0,0 +1,97 @@
...
+                Spacing = 3,
+                HorizontalContentAlignment = HorizontalAlignment.Right,
+                Items =
+                {
+                    saveButton,
+                    cancelButton,
+                }
+            });
+
+            Content = new Scrollable { Content = uiTextLayout };
```
WiIIiam278, line 94: Could Content be set to a two-row TableLayout, with the Save & Cancel buttons added to the second row and the scrollable added to the first? It'd be nice to be able to hit those buttons without needing to scroll to the bottom

src/SerialLoops/Dialogs/EditUiTextDialog.cs
```diff
@@ -0,0 +1,97 @@
...
+                    }
+                    else
+                    {
+                        _project.UiText.Messages[currentIndex] = uiTextBox.Text.GetOriginalString(_project);
+                    }
+                };
+
+                GroupBox uiTextGroup = new()
+                {
+                    Content = uiTextBox,
```
WiIIiam278, line 62: I think it'd be a really good idea to have some sort of predefined labels above each box to identify what each of these do. Some of them like `#P07` and the glyphs would be confusing for users, and these will probably be hard to edit without context.
The use of GroupBox here makes me think this may be planned :)?

src/SerialLoops/MainForm.eto.cs
```diff
@@ -198,7 +201,7 @@ private void InitializeProjectMenu()
             {
                 fileMenu.Items.AddRange(new[] { saveProjectCommand, projectSettingsCommand, migrateProjectCommand, exportPatchCommand, closeProjectCommand });
             }
-            Menu.Items.Add(new SubMenuItem { Text = "&Tools", Items = { renameItemCommand, searchProjectCommand, findOrphanedItemsCommand } });
+            Menu.Items.Add(new SubMenuItem { Text = "&Tools", Items = { renameItemCommand, editUiTextCommand, searchProjectCommand, findOrphanedItemsCommand } });
```
WiIIiam278, line 204: It might also be neat to add a button to this from the existing Project-specific settings menu (where you edit the banner data)

WiIIiam278 (approved):
This is neat -- though I do think perhaps could do with labels for each message box. Will approve to unblock for now though if you'd rather delegate some of this stuff to later PRs and just open issues :)

src/SerialLoops/Dialogs/EditUiTextDialog.cs
jonko0493: Yes that's a really brilliant idea. I'll do that.

src/SerialLoops/Dialogs/EditUiTextDialog.cs
jonko0493: Yes indeed, it is planned. :) But not in this PR, too much of a pain to break all these out manually.

src/SerialLoops/MainForm.eto.cs
jonko0493: Good idea, actually -- I'll see if I can do that.

jonko0493: On reflection -- I think this would lead to there being a dialog called from a dialog, and I'm not sure that's a good idea.

src/SerialLoops/Dialogs/EditUiTextDialog.cs
jonko0493: Jk, doing this now lol

PR #199, opened 2023-06-12
Fix no emulator path being set not being detected when trying to Build & Run

WiIIiam278:
The fact no emulator path has been set when trying to build and run is not currently being detected. Also added a thing to automatically redirect the user to the Preferences menu if this is the case.


jonko0493 (approved):
Great work!

PR #201, opened 2023-06-13
Add sprite and chibi GIF/sheet export

### LINKED ISSUE #170: Add sprite/chibi sheet/animation extraction
Has been requested and will be needed for #89 and #90 anyway.

Chokuretsu lib change: probably should try to find a better way to do this than relying on ffmpeg.

jonko0493:
Closes #170.

This adds buttons to the chibi and character sprite editors that will export the sprite/chibi as a GIF or as individual frames. The former is fun and the latter will be useful for editing these items once replacement is added.

### DIFF
--- a/src/SerialLoops/Editors/CharacterSpriteEditor.cs
+++ b/src/SerialLoops/Editors/CharacterSpriteEditor.cs
@@ -75,6 +75,7 @@ namespace SerialLoops.Editors
                             _log.LogError($"Failed to export mouth animation {i} for sprite {_sprite.DisplayName} to file: {ex.Message}\n\n{ex.StackTrace}");
                         }
                     }
+                    MessageBox.Show("Character sprite frames exported!!", "Success!", MessageBoxType.Information);
                 }
             };
 
@@ -109,7 +110,7 @@ namespace SerialLoops.Editors
                     }
 
                     LoopyProgressTracker tracker = new();
-                    _ = new ProgressDialog(() => frames.SaveGif(saveFileDialog.FileName, tracker), () => MessageBox.Show("GIF exported!"), tracker, "Exporting GIF...");
+                    _ = new ProgressDialog(() => frames.SaveGif(saveFileDialog.FileName, tracker), () => MessageBox.Show("GIF exported!", "Success!", MessageBoxType.Information), tracker, "Exporting GIF...");
                 }
             };
 
--- a/src/SerialLoops/Editors/CharacterSpriteEditor.cs
+++ b/src/SerialLoops/Editors/CharacterSpriteEditor.cs
@@ -1,8 +1,13 @@
 ﻿using Eto.Forms;
 using HaruhiChokuretsuLib.Util;
+using SerialLoops.Dialogs;
 using SerialLoops.Lib;
 using SerialLoops.Lib.Items;
 using SerialLoops.Utility;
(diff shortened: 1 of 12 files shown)


src/SerialLoops/Editors/ChibiEditor.cs
```diff
@@ -50,10 +54,59 @@ private Container GetBottomPanel()
...
+                    foreach ((SKGuiImage frame, int timing) in _animatedImage.FramesWithTimings)
+                    {
+                        for (int i = 0; i < timing; i++)
+                        {
+                            frames.Add(frame.SkBitmap);
+                        }
+                    }
+
+                    LoopyProgressTracker tracker = new();
+                    _ = new ProgressDialog(() => frames.SaveGif(saveFileDialog.FileName, tracker), () => MessageBox.Show("GIF exported!"), tracker, "Exporting GIF...");
```
WiIIiam278, line 82: `MessageBox.Show("GIF exported!")`

This produces a MessageBox with "Error" as the title -- should probably specify the title here to avoid a "task failed successfully" here

src/SerialLoops/Editors/CharacterSpriteEditor.cs
```diff
@@ -21,11 +26,109 @@ public override Container GetEditorPanel()
...
             _animatedImage.Play();
 
+            Button exportFramesButton = new() { Text = "Export Frames" };
+            exportFramesButton.Click += (sender, args) =>
+            {
+                SelectFolderDialog folderDialog = new()
+                {
+                    Title = "Select character sprite export folder"
+                };
+                if (folderDialog.ShowAndReportIfFolderSelected(this))
```
WiIIiam278, line 36: Since we're asking the user to select a folder here I think a MessageBox popup confirming that saving has occurred successfully would be good here; it's fine to not do this elsewhere since we're opening a select save file location dialog, but here it'd probably make sense to provide feedback as we're saving multiple files

src/SerialLoops/Editors/ChibiEditor.cs
```diff
@@ -100,9 +165,20 @@ private StackLayout GetFramesStack()
...
-                _framesStack.Items.Add(image);
+                StackLayout frameLayout = new()
+                {
+                    Orientation = Orientation.Vertical,
+                    Spacing = 3,
+                    HorizontalContentAlignment = HorizontalAlignment.Center,
+                    Items =
+                    {
+                        image,
+                        $"{timing} frames",
```
WiIIiam278, line 178: When I did the chibi UI here I did a lazy fix to display an invisible image where no frames were set for a direction of a chibi sprite.
We should probably fix this properly for this as now this displays "1 Frame" with no frame graphic above as a result, probably by disabling chibi button directions and defaulting to a direction that has present graphics if the animation type was changed and an invalid direction was selected.

WiIIiam278 (approved):
this is very fun, nice stuff! left a few comments

src/SerialLoops/Editors/ChibiEditor.cs
jonko0493: I'll see if I can do the buttons disabling thing in a sec, but for now just made the "1 Frames" thing go away

jonko0493: Done!

PR #204, opened 2023-06-15
Package improvements

### LINKED ISSUE #131: Add rpm package
We should add Fedora/RedHat RPM packages to our releases to better support those distros.

jonko0493:
Closes #131.

The main change here is adding the rpm package, but we also add a `Depends` line to the deb package spec that makes it so we don't have to specify "run this command to install OpenAL" anymore! It also gets installed with the rpm package, so we've really covered our bases here.


README.md
```diff
@@ -34,10 +34,7 @@
 * On macOS and Linux, run `sudo dkp-pacman -S nds-dev` from the terminal after installing the devkitPro pacman distribution.
 
 #### Installing OpenAL (Linux)
-If you're running on Linux, you will also need to install OpenAL, needed for audio processing. On Ubuntu/Debian (which are the distros we test on), it can be installed in a single command:
-```bash
-sudo apt install libopenal-dev
-```
+If you're running on Linux and _not using one of the package releases_ (the `.deb` or `.rpm`), you will also need to install OpenAL which is used for audio processing.
```
WiIIiam278, line 37: We should probably keep the `sudo` command, just put it in a <details> block

<details>
<summary>Install command</summary>

```bash
sudo apt install libopenal-dev
```
</details>

WiIIiam278 (approved):
Looks good :)

PR #213, opened 2023-06-16
Make chibis enter in the right order with enter/exit command

### LINKED ISSUE #141: Chibis don't display in the correct order when using CHIBIENTEREXIT command
The starting chibis can be arranged in any order, but `CHIBIENTEREXIT` causes chibis to line up and rearrange themselves in a particular order. The script preview should respect that order.

jonko0493:
Fixes #141.

Just a quick fix to have the chibi enter/exit command reflect what happens in the game.


WiIIiam278 (approved):
Looks good to me, yep

PR #214, opened 2023-06-18
Add the ability to apply hacks

### LINKED ISSUE #108: Add "apply hacks" feature
Add feature to apply pre-made ROM hacks to your project.

Example of common hacks:
* Expand the `SCENE_GOTO` array (this should maybe be done automatically)

jonko0493:
This adds a little "Apply Hacks" dialog that provides a list of common hacks that can be applied. After checking one or more off and hitting save, the hacks are assembled and inserted into the ARM9/overlays. This PR only implements a hack to skip the OP as it was simple to implement.

Closes #108.


src/SerialLoops/Dialogs/RomHacksDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+using System.Linq;
+using System.Text.Json;
+
+namespace SerialLoops.Dialogs
+{
+    public class RomHacksDialog : Dialog
+    {
+        private const int NUM_OVERLAYS = 26;
+
+        public RomHacksDialog(Project project, Config config, ILogger log)
```
WiIIiam278, line 20: Should set the `Title` property here so the window has a name

src/SerialLoops/Dialogs/PreferencesDialog.cs
```diff
@@ -69,7 +70,19 @@ private StackLayout InitializeOptions()
...
+                            },
+                            new BooleanOption
+                            {
+                                Name = "Use Docker for ASM Hacks",
+                                Value = Configuration.UseDocker,
+                                OnChange = (value) => Configuration.UseDocker = value
+                            },
+                            new TextOption
+                            {
+                                Name = "DevkitARM Docker Tag (for use with ASM Hacks)",
```
WiIIiam278, line 82: This long name and changing the sizes here unfortunately makes the preferences dialogue look pretty messy. Might want to do something to ensure all the boxes are equal in width. In the future (esp. with the key bindings feature request) it'd be good to perhaps add a list on the sidebar to let the user flick through multiple categories of preferences.

src/SerialLoops/MainForm.eto.cs
```diff
@@ -224,6 +226,9 @@ private void InitializeProjectMenu()
             closeProjectCommand.Executed += (sender, args) => CloseProjectView();
 
             // Tools
+            Command applyHacksCommand = new() { MenuText = "Apply Hacks" };
```
WiIIiam278, line 229: This -- and "Edit UI Text" (whoops, forgot to spot this earlier) -- should end with an ellipses (...) to indicate they open another dialog window. Some icons would also be nice, will prepare some for these new options.

src/SerialLoops/Dialogs/RomHacksDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+                    project.Settings.File.RomInfo.ARM9Ovt.First(o => o.Id == overlay.Id).RamSize = (uint)overlay.Length;
+                }
+
+                // We don't provide visible errors during the compilation of the hacks because it will deadlock the threads
+                // So at the end, we should check if any of the hacks that were supposed to be applied are not applied,
+                // and if there are some then we should let the user know.
+                IEnumerable<string> failedHackNames = appliedHacks.Where(h => !h.Applied(project)).Select(h => h.Name);
+                if (failedHackNames.Any())
+                {
+                    log.LogError($"Some hacks ({string.Join(", ", failedHackNames)}) failed to apply -- check logs to see why.");
```
WiIIiam278, line 186:
```suggestion
                    log.LogError($"Failed to apply the following hacks to the ROM:\n{string.Join(", ", failedHackNames)}\n\nPlease check the log file for more information.");
```

src/SerialLoops/Dialogs/RomHacksDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+
+            StackLayout buttonsLayout = new()
+            {
+                Orientation = Orientation.Horizontal,
+                Spacing = 3,
+                Items =
+                {
+                    importButton,
+                    cancelButton,
+                    saveButton,
```
WiIIiam278, line 200:
```suggestion
                    saveButton,
                    cancelButton,
```
Save and cancel are the wrong way round here -- just need swapping around.

src/SerialLoops/Dialogs/RomHacksDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+                    log.LogError($"Some hacks ({string.Join(", ", failedHackNames)}) failed to apply -- check logs to see why.");
+                }
+
+                Close();
+            };
+
+            StackLayout buttonsLayout = new()
+            {
+                Orientation = Orientation.Horizontal,
+                Spacing = 3,
```
WiIIiam278, line 195: Perhaps also add a bit of padding (~5?) -- the buttons here are a touch claustrophobic right up against the scrollable box

WiIIiam278 (approved):
Big fan of this -- especially the modular nature that lets users make their own hacks (might want to document that format). A few changes to make, but happy to unblock this tbh now that you've fixed the duplication issue.

I'll prepare some icons for this new stuff perhaps just before release. Only major bit of feedback is to do with the adjustments made to the preferences dialogue, but I think a quick fix for the box sizes would be fine for 0.2 :)

src/SerialLoops/Dialogs/RomHacksDialog.cs
```diff
@@ -0,0 +1,217 @@
...
+                }
+
+                // Get the overlays
+                List<Overlay> overlays = new();
+                string originalOverlaysDir = Path.Combine(project.BaseDirectory, "original", "overlay");
+                string romInfoPath = Path.Combine(project.BaseDirectory, "original", $"{project.Name}.xml");
+                string newRomInfoPath = Path.Combine(project.BaseDirectory, "rom", $"{project.Name}.xml");
+                foreach (string file in Directory.GetFiles(originalOverlaysDir))
+                {
+                    overlays.Add(new(file, romInfoPath));
```
WiIIiam278, line 148: Getting a crash here caused by a FileNotFoundException when checking the "Skip OP" hack and selecting "Save" before having ever applied it -- it's trying to find a file `SerialLoops\Projects\UITe\base\original\UITe.xml` in the basedir that doesn't exist (yet, I presume, that gets generated when applying, perhaps)

src/SerialLoops/Dialogs/PreferencesDialog.cs
jonko0493: Yeah that's smart. I've just shortened the title for now.

src/SerialLoops/Dialogs/RomHacksDialog.cs
jonko0493: Ah, it gets added when you create a new project. So this is an issue that's created by working on old projects. I'll see if I can add something that checks for its presence and copies it if it doesn't exist in the meantime.

PR #216, opened 2023-06-18
Add episode headers to script preview

### LINKED ISSUE #215: Add episode headers to script preview
Now that we have system textures, we should be able to add episode header previews when the `EPHEADER` command is called. We should do this.

jonko0493:
Closes #215.

Simple implementation, but adds the episode headers just as you'd expect! If an `EPHEADER` command is present, we skip drawing the rest of the top screen, which makes for greater efficiency as well.

image


WiIIiam278 (approved):
Looks good as is — we might want to consider our own warnings / checks against certain duplicate commands like EP_HEADER; even if it works it’s clearly not intended and can crash if there’s three or more (perhaps refusing to build until they’re addressed with optional override)

jonko0493:
We should definitely add some sort of script validation! That sounds like a super useful feature.

PR #220, opened 2023-06-19
Lock down project names, add save button to the toolbar, and add view logs menu item

### LINKED ISSUE #218: Spaces in project name mess up ASM hacks, other things probably
We should have validation/restrictions on project names to prevent this.

jonko0493:
Fixes #218.

* Project names now have a set of valid characters; attempting to type or paste invalid characters will result in input being canceled.
* Add a save button toolbar (feedback from user study)
* Add a view logs button to the Application/File menu (makes it easier to access the logs file; also feedback from user study). The button opens the SerialLoops.log file with the system associated program for viewing log files.

### DIFF
--- a/src/SerialLoops/Dialogs/ProjectCreationDialog.eto.cs
+++ b/src/SerialLoops/Dialogs/ProjectCreationDialog.eto.cs
@@ -22,8 +22,7 @@ namespace SerialLoops
         private Label _romPath;
 
         private const string NO_ROM_TEXT = "None Selected";
-        private const string ASSETS_URL = "https://github.com/haroohie-club/ChokuretsuTranslationAssets/archive/refs/heads/main.zip";
-        private const string STRINGS_URL = "https://github.com/haroohie-club/ChokuretsuTranslationStrings/archive/refs/heads/main.zip";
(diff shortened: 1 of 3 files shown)


src/SerialLoops/Dialogs/ProjectCreationDialog.eto.cs
```diff
@@ -34,7 +33,14 @@ void InitializeComponent()
             _nameBox = new()
             {
                 PlaceholderText = "Haroohie",
-                Size = new Size(150, 25)
+                Size = new Size(150, 25),
+            };
+            _nameBox.TextChanging += (sender, args) =>
+            {
+                if (args.NewText.Any(c => !ALLOWED_CHARACTERS.Contains(c)))
```
WiIIiam278, line 40: Perhaps a regex would be a lil easier here?
`[A-Za-z-_~.]+`

WiIIiam278 (approved):
Looks good :)

src/SerialLoops/Dialogs/ProjectCreationDialog.eto.cs
jonko0493: Regex will probably be simpler. I'll implement that change.

PR #228, opened 2023-06-20
Add ASM hack description to dialog, Makefile & hack updating, devkitARM version detection, and disabling of attract sequence for Skip OP hack

### LINKED ISSUE #224: Built-in "Skip OP" hack doesn't prevent the OP from playing during attract mode
It probably should, I think, given the purpose of this is to hide the OP since it can't be otherwise edited

### LINKED ISSUE #226: Hacks should have a "description" element that the UI should display
What it says on the tin.

### LINKED ISSUE #227: Upgrade Makefile to support devkitARM latest version (and add detection of outdated devkitARM)
devkitARM has updated and we should update with it so new installations work. Additionally, we should maybe try to detect when devkitARM needs to be updated to work with the latest version of Serial Loops.

jonko0493:
Fixes #224. Closes #226. Closes #227.

Does a couple of things because I got carried away lol.

* ASM hacks now have a description as part of their metadata and that gets displayed in the ASM hack dialog. This description is displayed when you mouse over a particular hack.
* Updated the Makefiles to work with the latest version of devkitARM and added some functionality to detect if the Makefiles are out of date compared to what the application has. Additionally, added functionality to determine if the devkitARM installation can be upgraded.
* Hacks can be updated now. If we add more files or change the description of a hack and push out an update, Serial Loops will detect this and upgrade the hack in the Hacks folder.
* Skip OP ASM hack now also disables the attract sequence timer on the main menu, preventing the OP from ever playing.

image

### DIFF
--- a/src/SerialLoops/Dialogs/AsmHacksDialog.cs
+++ b/src/SerialLoops/Dialogs/AsmHacksDialog.cs
@@ -1,4 +1,5 @@
-﻿using Eto.Forms;
+﻿using Eto.Drawing;
+using Eto.Forms;
 using HaroohieClub.NitroPacker.Patcher.Nitro;
 using HaroohieClub.NitroPacker.Patcher.Overlay;
 using HaruhiChokuretsuLib.Util;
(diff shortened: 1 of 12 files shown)


src/SerialLoops/Dialogs/AsmHacksDialog.cs
```diff
@@ -14,25 +15,41 @@
...
+                Size = new(300, 400),
             };
 
             foreach (AsmHack hack in config.Hacks)
             {
                 CheckBox hackCheckBox = new() { Text = hack.Name, Checked = hack.Applied(project) };
+                hackCheckBox.MouseEnter += (sender, args) =>
+                {
+                    descriptionLayout.Items.Clear();
+                    descriptionLayout.Items.Add(new Label { Text = hack.Name, Font = SystemFonts.Bold() });
```
WiIIiam278, line 50:
```suggestion
                    descriptionLayout.Items.Add(ControlGenerator.GetTextHeader(hack.Name));
```
Can probably use our existing styled header code here

src/SerialLoops/Dialogs/AsmHacksDialog.cs
```diff
@@ -231,7 +248,8 @@ public RomHacksDialog(Project project, Config config, ILogger log)
                     new Scrollable
```
WiIIiam278, line 248: Can we add a `{ Spacing = new(10, 0) }` to the main `TableLayout` here just to add a bit more spacing between the description panel and main hacks list scrollable?

WiIIiam278 (approved):
Nice work, this is a pretty needed change

PR #230, opened 2023-06-20
Fixup bad item renaming and command tree scrollbar bugs

### LINKED ISSUE #223: Editing a place or character item, then attempting to rename items causes items to be renamed as with the previous place/character name when saving while moving caret down the item explorer

### LINKED ISSUE #229: Moving the mouse into the script command panel while dragging the scrollbar starts moving commands
This is just because we use the `MouseMove` event and check that the primary button is down and the coordinates are over a command, but not whether the initial click _originated_ on a command. Can be a bit annoying for touchpad users especially who rely on the scrollbar more.

jonko0493:
Fixes #223. Fixes #229.

The rename bug was caused by the simple fact that our rename function was built assuming someone was doing right click -> rename (or F2), and thus renamed the currently selected item. I've abstracted the function out to allow providing a rename of an arbitrary item with the caveat that I don't know how to ensure that its name is then changed in the item selector tree.

The other bug has been _mostly_ fixed. There are some edge cases (specifically, if you pick up a command and drag it out of the box and drop it, the bug can still occur), but the vast majority of cases have been fixed.

### DIFF
--- a/src/SerialLoops/Controls/ScriptCommandSectionTree.cs
+++ b/src/SerialLoops/Controls/ScriptCommandSectionTree.cs
@@ -2,12 +2,12 @@
 using Eto.Drawing;
 using Eto.Forms;
 using HaruhiChokuretsuLib.Archive.Event;
+using HaruhiChokuretsuLib.Util;
+using SerialLoops.Editors;
 using SerialLoops.Lib.Script;
 using System;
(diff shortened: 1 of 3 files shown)


WiIIiam278 (approved):
Seems to do the job :) 👍

PR #251, opened 2023-06-21
Improve stability with many bugfixes

### LINKED ISSUE #236: Empty script sections get deleted on build/reload but labels for those sections persist
This results in duplicate labels which almost certainly are problematic, but even if they're not, they're very annoying

### LINKED ISSUE #243: Crash when attempting to replace BGM with an MP3
Confirmed to exist cross-plat (Mac & Windows). The error happens when attempting to cache the audio file:

```
System.IO.IOException: 'The process cannot access the file 'C:\Users\User\SerialLoops\Caches\bgm\BGM012.wav' because it is being used by another process.'
```

jonko0493:
Just a grab bag of bugs fixed through testing.

* Fixes quite a few errors with the script preview, most notably how sprites entrances/exits are handled (matches what the game does better)
* Uses display name for BGM selector in the script editor
* Topic links would cause crashes when clicking them before, this fixes that
* Fixes a crash on macOS when renaming without a selected command
* Fixes #236 by now pruning the labels section on load, preventing an issue where scripts would have lots of extraneous section references
* The BG_DISPCG command was initialized with a TEX_BG before; this has been fixed
* Fixes #243 where trying to replace a BGM with a file that needed AOT conversion caused crashes due to me doing a bad job copying them
* Some more character rename bugs were fixed
* Disable fields in the puzzle and character editors that can't currently be edited
* Fixes some errors with the Skip OP hack
* During investigation, I found that Docker does not work on macOS at all and requires elevation on Linux. I've updated the readme and disabled those options on macOS as a result
* We now allow loading projects even if your devkitARM installation is out of date bc that's fine as long as you're not trying to apply hacks
* Added melonDS snap location as a preset on Linux
* Fixed some underlying issues with NitroPacker
* Fixed a crash related to BG_DISP being set to NONE and prevented the user from selecting none for BG_DISP and BG_DISPCG

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,10 @@
 #### Installing Make or Docker
 To assemble ASM hacks you want to apply, you will need to decide whether to use Make or Docker.
 
+Currently, the Docker path is **only supported on Windows** due to operating system and framework limitations. It is possible to get Docker running
+just fine on Linux distros by running SerialLoops as root (e.g. `sudo SerialLoops`), but it's easier to just use Make. On macOS, there is no known
+way of getting the Docker path to work, so you will have to use Make.
+
 * [Make](https://www.gnu.org/software/make/) is the software used to assemble assembly hacks. Installing Make allows you to build the hacks
   directly on your system.
     - To install on Windows, you will have to use a terminal and a package manager. Your options are Winget (installed by default on Win10+) or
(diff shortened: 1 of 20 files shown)


WiIIiam278 (approved):
Rolling with this at the moment and it's working great, much more smooth. Nothing immediately jumps out

Issue #250, opened 2023-06-21
Add a palette editor


WiIIiam278:
It'd be nice to add a palette editor to allow for quick refinements and color replacements of images

Issue #249, opened 2023-06-21
Improve color image replacement of systexes with transparency


WiIIiam278:
Systexs with transparency are difficult to replace right now due to the lack of a way of changing their palette

PR #252, opened 2023-06-21
Add a palette viewer and "replace with palette" to the system texture editor

jonko0493:
This adds a palette viewer for systex items, the first step towards completing #250 . It also adds the ability to replace an image and provide a new palette, the first step toward #249 .


WiIIiam278 (approved):
I saw this and thought it'd be some monstrous complex thing and it's like 5 lines. I love it.

PR #254, opened 2023-06-22
Make it so you can replace system textures with transparent images to preserve transparency

jonko0493:
This one is really cool imo!


WiIIiam278 (approved):
oh this is really neat!

PR #255, opened 2023-06-22
Fixup subtitle preview

### LINKED ISSUE #242: Subtitle line preview requires modifying/updating text to get it to properly display
The preview for subtitled lines doesn't seem to appear by default, even if a subtitle has been set.

Changing the text's position gets the background to appear, but the text isn't drawn correctly; sometimes only punctuation characters are shown, and other times nothing at all (despite text being in the box)

jonko0493:
Fixes #242


WiIIiam278 (approved):
huh I thought I was running the update method there but ig not. Totally forgot to do the language conversion though

PR #258, opened 2023-06-22
Fixup system texture tiling replacement

### LINKED ISSUE #256: Only able to replace half of SYSTEX_EPISODE_SMALL
Any chosen image will only replace the left half of the image, with the right being white, for some reason -- possibly since the actual image is laid out vertically stacked?

jonko0493:
Fixes #256 

I stole the code for systex replacement from the BG replacement. BGs always tile 64x64, so that's hardcoded there; I thought I replaced it with the system texture TileWidth & TileHeight properties but apparently I did not. This prevents replacement from working fully as the tiles are too big.

Works now though. :)


WiIIiam278 (approved):
Works now, great.

PR #264, opened 2023-06-22
Add OP_MODE chibi change hack

jonko0493:
Adds a hack to change the chibi that walks out when using the `OP_MODE` command.


WiIIiam278 (approved):
image

Awesome :)

PR #287, opened 2023-06-24
Bugfixes & user study feedback

### LINKED ISSUE #271: Search loading bars appear below the advanced search window
This is because the search window is locally `Topmost`. Perhaps we should try setting topmost on the loading dialogs.

### LINKED ISSUE #274: Sprite shakes must affect sprite positioning
Currently, they don't, but the sprite shake actually moves the sprite (e.g. shake right will move a centered sprite to the right).

### LINKED ISSUE #275: Canceling out of long BGM replacement and then trying to replace another BGM causes crash
We should add a cancellation token to long replacements and cancel them if the user attempts to do another replacement, as well as put a try/catch around this.

### LINKED ISSUE #276: Text missing from the saving dialog on adjust volume
Should add that so it's clear what's happening.

### LINKED ISSUE #277: Script editor numeric steppers for things with frame counts can go negative
We should clamp them to zero so you can't put negative numbers in as that would definitely lead to overflow behavior.

### LINKED ISSUE #281: Attempting to open an empty script causes crash
Delete all commands from a script item and then attempt to reopen it without adding anything. SL will crash attempting to resolve the commands in the script item.

### LINKED ISSUE #283: Significant errors when adding dialogue commands sometimes
Looks to be freaking out when attempting to parse the voice font parameter.

jonko0493:
Fixes #271. Fixes #274. Fixes #275. Fixes #276. Fixes #277. Fixes #281. Fixes #283.

A bunch of bugfixes and some tweaks to address user study feedback.

In addition to the linked issues, this also attempts to address some other issues with sprite entrance effect previewing to match what happens in-game.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,10 @@ Documentation for how to use Serial Loops can be found on [our website](https://
 
 ## Installation
 ### Prerequisites
-#### Installing devKitARM
+#### A Nintendo DS Emulator
+To test the game easily, you will want to have a Nintendo DS emulator installed. We recommend using [melonDS](https://melonds.kuribo64.net/) for its accuracy.
+
+#### Installing devkitARM
 [devkitARM](https://devkitpro.org/wiki/Getting_Started) is required to use Serial Loops on all platforms.
 
 * Using the Windows graphical installer, you can simply select the devkitARM (Nintendo DS) workloads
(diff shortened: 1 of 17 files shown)


WiIIiam278 (changes requested):
Tested each of them and works on win :)

WiIIiam278 (approved):

PR #293, opened 2023-06-26
Add SFX items & Character voice font/text timer preview/editing

### LINKED ISSUE #212: Add SFX items
Requires porting code over from NitroStudio to be able to play/replace SFX from the SDAT (snd.bin). Also will require some work to figure out exactly how the SFX indices reference them and how sound banks are loaded, etc.

### LINKED ISSUE #245: Make voice font and text timer editable
Pending #93, we need to make the voice font and text timer editable. Since there will be a preview for the voice font, we should also consider implementing a preview for the text timer.

jonko0493:
Closes #212. Closes #245.

This adds SFX items (but does not make them editable). It allows for previewing them with (unfortunately) a new sound player that uses modified versions of Gota7's SDAT tech used in Nitro Studio 2. I also added the ability to export a wav of an SFX using the same tech.

Additionally, this shows the _bank_ that the SFX uses, and the _groups_ which contain the SFX. The groups in particular are important as they are what determine which sounds are playable in a given scene. It's unknown where the game decides which groups to load for a given scene at this time.

Finally, now that we can preview the voice font, we allow specifying a different sound effect to use and allow changing the text timer as well.

Oh, and we bump the versions to 0.3 preview now. :)


WiIIiam278 (approved):
Works flawlessly -- excellent job! ~~(barring fixing up those tests)~~

PR #295, opened 2023-06-27
Fixup corrupted audio

### LINKED ISSUE #268: BGM020 plays corrupted audio
Converts out just fine, all the other BGMs work fine, just when pressing the play button we get corrupted audio.

jonko0493:
Fixes #268.

This was my fault! I made it so we manually turn off looping in BGM audio to make it so we can export wavs, but it ironically also manually turned _on_ looping in ADX audio without loops, corrupting playback. This fixes that oversight.


WiIIiam278 (approved):
Didn't even notice this before, so good catch - plays fine :)

PR #296, opened 2023-06-27
Improve chibi loading perf

### LINKED ISSUE #269: Improve chibi loading perf
Chibi loading is really slow right now. Maybe there's something we can do at the lib level that will speed it up?

jonko0493:
Fixes #269.

This fix takes an underlying bugfix in the lib and gets rid of the very slow hack I put in to get around that limitation. Super easy, works beautifully.


WiIIiam278 (approved):
Amazing stuff -- big improvement :)

Load time on main:
image

Loading time on this branch:
image

PR #297, opened 2023-06-27
SerialLoops.Mac.app -> Serial Loops.app

### LINKED ISSUE #244: SerialLoops.Mac -> SerialLoops
The assembly name is correct, so it's some other property (probably in the DMG creation task) that's naming the app wrong. We should fix that.

jonko0493:
Fixes #244 and makes things look a lot nicer on Mac.


WiIIiam278 (approved):
Just built on macOS -- looks great!

PR #298, opened 2023-06-27
Add tutorial items

### LINKED ISSUE #210: Add tutorial items
Tutorial items are actually pretty simple, just an ID and an associated event. Might look into it further to see if there's more dials or knobs we can turn, but that's basically it.

~If they really are as simple as we think, perhaps we should just add them as information on the script files they're associated with (i.e. this is tutorial 1012, etc.).~ This would make them difficult to edit; reverting this to adding them as items.

jonko0493:
Closes #210.

This one was easy. While I was here, I also removed characters from the orphaned items dialog exclusion list and added references to them.


WiIIiam278 (approved):
Given how simple these are, I am _slightly_ skeptical of adding them as items as opposed to just inserting them into the script file -- but it is true that this would make these harder to move around.

I do wonder perhaps in the future if this might be better served by a simple "edit tutorial mappings" dialog/UI somewhere tucked away in the tools menu (perhaps with links to said menu from the scripts)?

After all, technically these items aren't really where you edit the tutorial itself; that's done in the individual scripts and BG textures. &mdash; and where possible, I do think we should avoid bulking out the item explorer in a way that may cause confusion

Just something to consider, I suppose :). 

This PR is of course totally fine for now and I think this is something we could reconsider, but i'll leave it up to you.

PR #300, opened 2023-06-29
Add chibi replacement

### LINKED ISSUE #90: Add chibi replacement
Chibis should be able to be replaced (and eventually added). 

We should be able to provide a series of images and timings and replace a chibi. Additionally, we should be able add not only new chibis, but new poses & animations for existing chibis and name them correctly so they work with what the game expects.

Lib changes: same changes as #89.

jonko0493:
This does the trick! The UI isn't great but we are able to replace chibis fully now!

Closes #90.


WiIIiam278 (approved):
Great work on this -- will look at the UI when I get a chance~!

PR #302, opened 2023-06-30
Fixup the Mac build frfr

jonko0493:
Build succeeded so it actually works this time.


WiIIiam278 (approved):
Unblocking this, but I'll also test building on Mac in a little while

PR #303, opened 2023-09-17
Add episode & episode unique search queries

jonko0493:
Adds some search queries for finding all items that appear in a specific episode and that are unique to a specific episode.

### DIFF
--- a/src/SerialLoops.Lib/SearchQuery.cs
+++ b/src/SerialLoops.Lib/SearchQuery.cs
@@ -22,7 +22,9 @@ public class SearchQuery
         Script_Flag = 12,
         Speaker_Name = 13,
         Conditional = 14,
-        Background_Type = 15
+        Background_Type = 15,
+        Episode = 16,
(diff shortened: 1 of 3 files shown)


src/SerialLoops.Lib/SearchQuery.cs
```diff
@@ -22,7 +22,9 @@ public enum DataHolder
         Script_Flag = 12,
         Speaker_Name = 13,
         Conditional = 14,
-        Background_Type = 15
+        Background_Type = 15,
+        Episode = 16,
```
WiIIiam278, line 26: Maybe "Episode_Number"? just to indicate that's what the user needs to type a lil more clearly

WiIIiam278 (approved):
very handy stuff

src/SerialLoops.Lib/SearchQuery.cs
jonko0493: makes sense to me!

PR #314, opened 2024-01-11
ASM hack QoL changes & upgrade to .NET 8.0

### LINKED ISSUE #306: Display logging when ASM patching
Currently, we just pop a terminal and begin patching. It would be nice to display what's happening in the patcher window or somewhere similar so that the user knows what's going on.

### LINKED ISSUE #307: Clean up Docker containers when we're done ASM patching
Currently we leave tons of docker containers just lying around after patching using Docker. We should clean up after ourselves with docker rm.

### LINKED ISSUE #308: src directory left in an unclean state on ASM patch failure
When we fail to apply an ASM hack, we still leave the source and build files in the `src` directory. These should be cleaned up to make sure that directory is tidy and application of future hacks doesn't fail (as without this cleanup it will be in a perpetual fail state).

### LINKED ISSUE #311: Upgrade to .NET 8.0
It's here, we should do it.

jonko0493:
Closes #306. Closes #307. Fixes #308. Closes #311.

The switch to .NET 8.0 makes this change look bigger than it is. While upgrading the choku lib to .NET 8, I made a bunch of underlying breaking changes to its APIs (just renaming classes and such) which is why you see a bunch of changes throughout.

Highlights:
* .NET 8 upgrade, including a bunch of small changes to code style that come with that upgrade to a newer C#
* Upgrade a ton of packages along the way, including bumping NitroPacker and the Choku lib to their .NET 8 versions
* Before, if a patch failed to apply, we left the build directory in an unclean state (with patch files just sitting there). New code explicitly reverts those patches and deletes unnecessary directories.
* Added functionality to NitroPacker to allow naming the docker containers we create when patching, which gives us the handy ability to call `docker rm` afterwards and clean up those containers
* Surfaced the make/docker logging to the user through a loading modal which will hopefully make that process less confusing

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Please file bugs in the Issues tab in this repository. Please include the follow
 Serial Loops is licensed under the GPLv3. See [LICENSE](LICENSE) for more information.
 
 ### Building
-Serial Loops requires the .NET 6.0 SDK to build. You can download it [here](https://dotnet.microsoft.com/download/dotnet/6.0). To build Serial Loops for your platform, run:
+Serial Loops requires the .NET 8.0 SDK to build. You can download it [here](https://dotnet.microsoft.com/download/dotnet/8.0). To build Serial Loops for your platform, run:
 
 ```bash
 dotnet build src/PLATFORM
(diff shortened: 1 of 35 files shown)


WiIIiam278 (approved):
Tested standard test ASM hack application (non-dockerized) on my mac, and it seems to work just fine (after an IDE upgrade to a version that supports 8!). After that, no friction!

I'll give Docker testing a go at some point soon - I love me a nice loading modal!

Issue #83, opened 2023-04-03
About dialog crashes if opened after closing


jonko0493:
* On Windows, the app crashes immediately after attempting to load the AboutDialog for the second time
* On macOS, the app loads the second AboutDialog, but then after closing it acts as if it's still open and refuses to allow interaction
* Works as expected on Linux

jonko0493:
The issue here is that the `Dialog.Close` method in WPF disposes the dialog, so it can't be reopened after closing it. (Something similar is likely true in MonoMac.) The solution is to new up an `AboutDialog` each time.

PR #316, opened 2024-01-12
Add merge pipeline for UI testing on Windows

### LINKED ISSUE #312: Onboard to Appium for Windows UI unit testing
We should unit test our UI, probably. Quite a few issues arise from UI changes that would probably be easier caught were we to unit test out stuff.

There's a few options for doing so; namely, just testing Wpf and using one of several available Wpf unit test libraries, or exploring something like appium [1] which provides testing drivers for multiple platforms.

[1] https://github.com/appium/appium-windows-driver

jonko0493:
Closes #312.

Based on the same branch as the other PR I have open right now. That one should merge first and then this one's changes will become more comprehensible.

It's possible to run these tests locally, but they require a bit of setup. There is some special casing introduced specifically to ensure the tests will run both locally and on the agent.

The two UI tests introduced in this PR are:
* A test designed to catch an unlikely reintroduction of #83. It's a super simple test that just opens/closes the about dialog twice.
* A scenario test for ASM hack application which tests both applying and reverting a single hack (notably, this hack only applies to the overlays -- perhaps we should consider expanding this test/adding a new test to test an ARM9 hack as well).


WiIIiam278 (approved):
Looks good as far as I can tell so far.

I don't have a Windows machine at the moment unfortunately so can't test this locally - but would certainly be intrigues to give this ago writing some tests for mac.

It's a slight shame that it _seems_ like we'll have to write platform-specific tests, but obviously in the case of the examples you illustrated they are rather platform specific. With that said, they seem quite nice to write (looks like it's quite trivial to select stuff by name attribute, etc). 

We should probably add some light documentation on how to execute tests and what that requires to README or the docs.

jonko0493:
Adding documentation on test running to the README makes sense to me. I don't think it's important to an end-user, so I would leave it out of the docs.

PR #318, opened 2024-01-13
Add Mac UI tests with Appium

### LINKED ISSUE #315: Onboard to Appium for Mac UI testing
Follow-up to #312 but for Mac.

We should see if we can get UI tests running on the hosted machines while bootstrapping devkitARM (Mac). If we can't, we could look into a self-hosted agent, but that might also be a pain. Unlike with Windows, we don't need to run Docker or WSL so there's no real reason that we should *need* a self-hosted agent that I can think of right now, unless devkitARM bootstrapping proves to be particularly difficult.

jonko0493:
Closes #315.

Tested and working locally and on Azure DevOps. ~Haven't written the ASM hacks one for this yet so not quite parity with the Windows tests, but that fine for now, probably?~ Wrote them lol

The screenshots that this one produces are way more useful than the ones that Windows produces, so that's also nice (unfortunate for Windows tho).

This also introduces a slight change in the main app's behavior as well: when looking for the devkitARM path on new install, we first check for the `DEVITKARM` environment variable (since this gets set on install). We then check that this path exists (as on Windows that variable can be set to `/opt/devkitpro/devkitARM`) and then proceed with the old behavior if it doesn't. This allows us to set up bootstrapped devkitARM on the mac agents since we can't install things to `/opt`.

Oh, while I'm here, the official build broke last night due to the net8.0 upgrade, so this fixes that.


jonko0493:
Alright, added some ASM hack tests, so yay! We have functional UI tests for both Windows and Mac now.

WiIIiam278 (approved):

PR #320, opened 2024-01-14
Update NitroPacker to allow for better Docker container deletion

### LINKED ISSUE #323: KBG preview is broken
Probably happened somewhere in the .NET 8 upgrade of the choku lib.

image

jonko0493:
Rather than assigning container names and deleting them at our level, I've made it so NitroPacker just passes the `--rm` parameter by default so containers get auto-removed after running.

Also fixes #323 by taking an update from the Chokuretsu lib that fixed the underlying problem.


WiIIiam278 (approved):
lol nice

Issue #329, opened 2024-01-18
Crop & scale dialog is not really functional on macOS


jonko0493:
* The crop area is not responsive to mouse panning
* Holding shift to pan the asset being scaled also does not work

jonko0493:
Currently it's not functional at all on the Avalonia branch. But once it is, this will be easily closeable.

cc/ @WiIIiam278

Issue #326, opened 2024-01-15
Right-clicking tabs has inconsistent behaviour between platforms


jonko0493:
* The area where you can right click to bring up the tab context menu is anywhere on Windows, but only on the Tab on mac
* You can't seem to Control+Right-click a TAB control on Mac

What probably needs to happen is us looking at Eto internals to fix this control up a bit.

WiIIiam278:
![Image](https://github.com/haroohie-club/SerialLoops/assets/31187453/6bfe75c0-f53c-4fdd-9b2c-32e5bb35a7f2)

Having difficulty reproducing

WiIIiam278:
Seems like Control+Clicking doesn't work, but right-click does.

After discussion, marking as blocked for now as this really needs some Eto changes to properly resolve.

jonko0493:
Closing because this is fixed in the Avalonia branch, at least as far as we're concerned.

Issue #322, opened 2024-01-14
Write some UI tests


jonko0493:
We're gonna need UI tests, which will be especially useful when we do things like #321.

We will need to make a TestAssets.zip file and download that onto the machines to provide images/scripts/etc. to compare against the ones we create during the tests.

Some ideas:
* Backgrounds
  - [x] Import a background from a test asset, re-export and see if it's the same -- the files themselves will be different, but appium has a `GetImagesSimilarity` method that would likely work for this
  - [x] The crop/resize tool should also be tested somehow
  - [ ] CG titles should be tested to see if they persist / end up in EXTRA.S
* BGMs
  - [ ] Have test asset audio and a waveform image of that audio, replace the audio and generate a waveform image, then compare the two images for similarity
    - Just having pre-gen'd audio waveforms for all our sounds and comparing against them would be a great unit test, actually. Worth thinking about.
  - [x] Check that extract produces a file of non-zero size, I suppose. Could do the same thing with replace.
  - [x] One thing that would work is Extract -> Replace -> Restore -> Extract resulting in two extracted files of the same size/content, so we could test that workflow and hit a couple of scenarios
  - [x] Not sure if there's a way we could test that the play/stop buttons work? I'm not sure how to check if sound is playing on a test machine (appium certainly doesn't have a built in method to do that) -- ~[this StackOverflow answer](https://stackoverflow.com/questions/23829571/how-to-know-if-a-sound-is-playing-using-c) implies this could be done through a library called CSCore. Worth looking into.~ Looked into it, this library is ancient. The way to do this on Windows is with NAudio's WASAPI implementation, which I've done, but it's very difficult to do it on macOS so I've ignored that problem for now.
  - [ ] Check that changing the BGM title persists (change it, close the tab, reopen the tab, see that the text is still the same). We could also check for the presence of EXTRA.S on disk after this change is made (and since we're designing from scratch, we could even compare it against a test asset or static string to see if the change propagated)
  - [ ] Probably ought to do something about loop properties and volume changing, but I'm not sure exactly how to go about it. Maybe waveform comparisons against test assets, but not sure.
* Character Sprites
  - [ ] Check a couple of them to see that both frames and gifs export properly (compare against test assets)
* Characters
  - [ ] Check that changing the character name changes the tab name
  - [ ] Check at least a few of the default colors to see if the color on the screen matches appropriately (could try to do this with screenshots but might only work on mac)
[...]

PR #331, opened 2024-01-18
Add more UI tests

### LINKED ISSUE #330: Background music doesn't stop playing when tab is exited via context menu
The culprit is the fact that the close tab command from that context menu just calls
```csharp
_tabs.Tabs.Remove(_tabs.Tabs.SelectedPage);
```
which doesn't trigger the `PageClosed` event on the `DocumentControl`.

Discovered via the UI tests!

jonko0493:
First bit of work on #322.

This implements five of the bullet points from that issue across a total of three tests for Windows and two tests for Mac.

I'll probably add on to this while I wait for #329 which is causing BG tests to fail on macOS (and maybe #326 which makes it so we can't close tabs on that platform either). Because of that, I'm drafting for now.


jonko0493:
Since we're switching to Avalonia, this is going to be out of date no matter what. Closing.

Issue #334, opened 2024-01-24
Factor in EVTTBL.S when dealing with script files


jonko0493:
This file determines the "read flags" that allow for dialogue skipping and such to work. We should make sure we're editing it as people make edits to the script files as well.

PR #336, opened 2024-01-27
Add a save editor (round 1)

### LINKED ISSUE #333: Add save editor tool
I just think it would be neat to have one of these lol.

jonko0493:
Closes #333.

This PR adds a new dialog accessible from the tools menu that allows for editing Chokuretsu save files. A lot of RE work had to be done to understand the save files better, so there were some changes to the Chokuretsu lib that propagated up here as well.

Currently, the (rather bad) UX flow works as follows: Tools -> Edit Save File prompts the user to select a save file through an open file dialog. The user is then presented with this screen:

image

These four buttons each represent different parts of the save file. Clicking any of them will open a second dialog that takes a long time to load.

image

The common save file dialog looks like this. The user is presented with common flags that exist across all save files in addition to options from the config menu.

image

The first two saves are checkpoint saves and their dialog looks like this. The time of the save can be adjusted as well the episode number that's displayed in the save. Finally, the index of the scenario command and the friendship levels can be edited.

image

Finally, this is the quick save slot dialog. In addition to the checkpoint save options, the current script, the current script block, and current script command index can be edited. This results in changing the saved script preview as well.

Future PRs should improve the UX for this component, especially with regard to load times. After the UX improvements are made, we should also introduce UI testing for this component. Additionally, more RE work should be done to better understand the remaining unknown components of the save file sections and mark down more of the flags. #334 will particularly affect our understanding of the "read flags" for script files.


WiIIiam278:
Testing this - amazing work so far. Truly great stuff! There's nitpicks I have with the UI, but I think it'd be easier to start a UI PR to go through and update some of those myself :)

src/SerialLoops.Lib/Flags.cs
```diff
@@ -0,0 +1,81 @@
...
+using SerialLoops.Lib.Items;
+using SerialLoops.Lib.Util;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SerialLoops.Lib
+{
+    public class Flags
+    {
+        private static readonly Dictionary<int, string> _names = new()
```
WiIIiam278, line 11: I think we perhaps ought to put these in a JSON resource rather than hardcode them in the logic here :)

jonko0493:
I agree -- the UI needs an overhaul rather than nitpicks haha

src/SerialLoops/Dialogs/SaveEditorDialog.cs
```diff
@@ -0,0 +1,121 @@
...
+                SaveSlotEditorDialog saveSlotEditorDialog;
+                if (slot is QuickSaveSlotData quickSave)
+                {
+                    saveSlotEditorDialog = new(_log, quickSave, _project, _tabs);
+                }
+                else
+                {
+                    saveSlotEditorDialog = new(_log, slot, _project, _tabs);
+                }
+                saveSlotEditorDialog.Show();
```
WiIIiam278, line 116: I think we should wrap this here with a ProgressTracker and offload this from the app thread as it's pretty slow.

src/SerialLoops/Dialogs/SaveEditorDialog.cs
```diff
@@ -0,0 +1,121 @@
...
+        void InitializeComponent()
+        {
+            Title = $"Save Editor - {Path.GetFileName(_saveLoc)}";
+            Width = 400;
+            Height = 400;
+            Button saveCommonButton = new() { Text = "Common Save Data", Height = 64 };
+            saveCommonButton.Click += (sender, args) =>
+            {
+                SaveSlotEditorDialog saveSlotEditorDialog = new(_log, _save.CommonData, _project, _tabs);
+                saveSlotEditorDialog.Show();
```
WiIIiam278, line 51: Same as with the other buttons, should prob offload this and track loading with a ProgressDialog if we can

src/SerialLoops/Dialogs/SaveEditorDialog.cs
```diff
@@ -0,0 +1,121 @@
...
+                    new(GetSaveSlotPreviewButton(_save.QuickSaveSlot)),
+                    new(),
+                    new(buttonsLayout),
+                },
+            };
+        }
+
+        private Button GetSaveSlotPreviewButton(SaveSlotData slot)
+        {
+            Button slotButton = new() { Text = $"EPISODE: {slot.EpisodeNumber}\n\n{slot.SaveTime:yyyy/MM/dd H:mm:ss}", Height = 64 };
```
WiIIiam278, line 104: We should call a method here to detect IsNewSave (would require a lib change) and display the "Create New Save" text as the button label here if the slot is unoccupied.

WiIIiam278 (changes requested):
A few smaller things here and there.

With regard to the flow here (and this is more UI stuff so can be delegated): I think it'd be nice if we could:
- move the Save Editor to the "File" menu (since it's an action involving a file, rather than a tool within project editing)
- which would allow us to expose this as an option in the main menu. It should then prompt the user to make a project (and then automatically open the Save Editor when done).
 
I think Save Editing is a fairly key use case (albeit not our primary goal) for Serial Loops and should be available from the top menu, and so we should make this easy to access.

src/SerialLoops.Lib/Flags.cs
jonko0493: Yeah, that's smart, especially for localization!

src/SerialLoops/Dialogs/SaveEditorDialog.cs
jonko0493: Yes, absolutely. I think the only reason I didn't was because I was hoping we'd change it so it didn't take as long. :P but no matter what it's gonna take a while so we should wrap it for now for sure.

src/SerialLoops/Dialogs/SaveEditorDialog.cs
jonko0493: Makes sense!

src/SerialLoops/Dialogs/SaveEditorDialog.cs
jonko0493: Yeah, lemme take a look at that scenario and see how best to tackle it!

jonko0493: Not sure a lib change is entirely necessary here! Looks like we could just detect "episode 0" & a zeroed out date and assume that means "new save". 

image

WiIIiam278: > Not sure a lib change is entirely necessary here! Looks like we could just detect "episode 0" & a zeroed out date and assume that means "new save".
> 
> image

Ah, alright then - sounds good!

jonko0493:
> move the Save Editor to the "File" menu (since it's an action involving a file, rather than a tool within project editing)
> which would allow us to expose this as an option in the main menu. It should then prompt the user to make a project (and then automatically open the Save Editor when done).

Regarding these two points -- do you mind if you handle these in your UI overhaul for this component? It is a pretty substantial flow break, I feel. I can definitely do it here if you would prefer that, of course!

WiIIiam278:
> > move the Save Editor to the "File" menu (since it's an action involving a file, rather than a tool within project editing)
> > which would allow us to expose this as an option in the main menu. It should then prompt the user to make a project (and then automatically open the Save Editor when done).
> 
> Regarding these two points -- do you mind if you handle these in your UI overhaul for this component? It is a pretty substantial flow break, I feel. I can definitely do it here if you would prefer that, of course!

Fine by me :+1: - just wanted to guage thoughts

src/SerialLoops/Dialogs/SaveEditorDialog.cs
jonko0493: Can't async this bc UI thread; will handle in UI overhaul

WiIIiam278 (approved):

PR #338, opened 2024-01-28
Add item items

### LINKED ISSUE #235: Add Item items
~These are the guys defined by `ITEM.S` in dat.bin and used via the `ITEM` script command. They display an image on the screen, but the exact specifics as to what to do with that image or how any of it works is very much a mystery. We should look more into the disassembly and see if we can figure out what they do and if they're worth adding to SL (even though they are unused in the base game).~

It's been figured out how they work and imo seems worth adding them so that they can be used if people want to do so.

jonko0493:
Closes #235.

This fully adds item items to SL -- it adds an item editor (allowing them to be replaced), adds new script parameters for items (controlling their location and transition), updates the `ITEM_DISPIMG` command in the script editor to use these new script parameters and an image selector for the item itself, and adds them to the script preview.

image

image


jonko0493:
Only thing left to do here is to add an Item icon by @WiIIiam278!

src/SerialLoops/Editors/ScriptEditor.cs
```diff
@@ -1998,6 +2024,35 @@ private void UpdatePreview()
                         canvas.DrawBitmap(spriteBitmap, sprite.Positioning.GetSpritePosition(spriteBitmap), sprite.PalEffect);
                     }
 
+                    // Draw items
+                    ScriptItemCommand lastItemCommand = commands.LastOrDefault(c => c.Verb == CommandVerb.ITEM_DISPIMG);
```
WiIIiam278, line 2028: Can we move this routine above the character sprite drawing? In-game, Items render behind characters (I think)

image

versus in game...

image

WiIIiam278 (changes requested):

WiIIiam278 (approved):

PR #339, opened 2024-01-29
Improve save editing UI

WiIIiam278:
Improves the Save Editing UI and dramatically ~~increases~~DE-creases (!) save load times.

This PR features save slot previews, a tabbed interface for editing saves, and pagination/filtering for setting flags.

<details>
<summary>Screeenshots</summary>

image
image
image
image
image

</details>


jonko0493:
hopefully it dramatically decreases the load times :P

src/SerialLoops/Dialogs/SaveSlotEditorDialog.cs
```diff
@@ -499,8 +212,563 @@ private void InitializeComponent()
...
+            };
+            flagPageBox.ValueChanged += (sender, args) =>
+            {
+                _flagPage = flagPageBox.Value = Math.Max(1, Math.Min(totalPages, flagPageBox.Value));
+                lastPageButton.Enabled = _flagPage > 1;
+                nextPageButton.Enabled = _flagPage < totalPages;
+                UpdateFlagList(_flagPage);
+            };
+
+            SearchBox filterBox = new() { PlaceholderText = "Filter by Name", Width = 200 };
```
jonko0493, line 246: It would be nice if pressing Enter in the filter box had the same effect as clicking the apply filters button.

jonko0493 (approved):
Looks really good. I've left feedback that should be addressed, but approving to unblock you either way.

PR #340, opened 2024-01-30
Add save clearing

jonko0493:
Also fixes a fun little bug with the previews I found! That I also introduced!


WiIIiam278 (approved):
nice work, looks good. Lol @ me still missing some curly braces. I really should look at configuring the rider config and completing that issue

PR #342, opened 2024-02-05
Add localization capabilities

jonko0493:
This basically redoes how we handle putting strings in the program. There's still some work to be done, particularly around localizing error messages coming from the lib, but on the whole it seems like it will work pretty well.


src/SerialLoops/Dialogs/PreferencesDialog.cs
```diff
@@ -117,16 +118,20 @@ private StackLayout InitializeOptions()
                             new DropDownOption(
                                 [
                                     ("en-US", "English (United States)"),
+                                    ("de", "Deutsch"),
                                     ("en-GB", "English (United Kingdom)"),
```
WiIIiam278, line 122: this took me awhile to figure out why you did this but I'm guessing the default lang first, then sort alphabetically. makes sense

jonko0493: Yeah that's exactly what it is. Lemme leave a comment

jonko0493: actually, i should probably just sort this alpha by language name lol

PR #345, opened 2024-02-07
Language -> Language Template in project creation

jonko0493:
From a conversation with @FuyukoAyumu, this seems to make more sense for what it actually functionally does.


WiIIiam278 (approved):
Yeah -- good idea.

PR #347, opened 2024-03-06
Improve perf with some concurrency magic

### LINKED ISSUE #310: Replace .First(idx => Index == idx) lookups with direct indexing
In the chokuretsu lib, there's currently a lot of places where we look up items by comparing their index to the lookup value via LINQ. This is slower than indexing directly and virtually always the index pattern is `rawIndex + 1`. If we direct index, we will likely see significant perf gains with no real loss in functionality.

jonko0493:
Closes #310 in both letter and spirit.

Introduces two new `ArchiveFile` methods from a lib change:
* `GetFileByIndex` &ndash; gets a file with direct indexing which is a huge perf bump from before
* `GetFileByName` &ndash; gets a file with by name with a PLINQ query which is faster than just doing it regularly

I made sure to replace the old file lookups we had with these two. Things should load quicker in general now.

In addition, in an effort to improve project load times, I removed the `ScriptUses` property from all items that have it and stopped calculating script usages at load/save time. Instead, we now calculate them as part of reference lookup. It turns out that this is preferable as load times are now blazing fast and you can't really tell that reference lookups have slowed down at all except for the orphaned files dialog which still isn't that slow (and is a lower priority than opening a project anyway).

### DIFF
--- a/src/SerialLoops.Lib/Build.cs
+++ b/src/SerialLoops.Lib/Build.cs
@@ -224,7 +224,7 @@ namespace SerialLoops.Lib
         {
             try
             {
-                GraphicsFile grpFile = grp.Files.FirstOrDefault(f => f.Index == index);
+                GraphicsFile grpFile = grp.GetFileByIndex(index);
 
                 if (index == 0xE50)
                 {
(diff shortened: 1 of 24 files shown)


jonko0493:
Test failures are legit; will need timet o investigate them.

WiIIiam278 (approved):
Oh yeah this is definitely a lot faster :) nice stuff!

PR #348, opened 2024-06-20
Bump package versions

jonko0493:
Just a quick chore to bump the package versions of a lot of things we use.


WiIIiam278 (approved):
All fine so long as CI passes in this case :)

PR #354, opened 2024-07-15
Add character sprite editor, remove SLG maps as options from ADV mode

### LINKED ISSUE #89: Add character sprite replacement
We should be able to replace character sprites (and eventually add new ones). The process would look something like this:

* The user prepares a base pose, a set of four to eight images comprising the eye animation, and a set of four to eight images comprising the mouth animation.
* The user selects the sprite they want to do replacement on. They are then allowed to select between all of the existing poses, eye animations, and mouth animations. At this point, they may opt to replace some of these graphics with the ones they have uploaded. Alternatively, they should be able to add new graphics to the archive to use.
  - When replacing/adding mouth/eye animations, they must specify the timings for each animation frame (which they will do with the help of a preview)
* After selecting the three sprite components they want for the particular sprite they're replacing, the user can then arrange the sprite components on the screen to create a completed image.

Lib changes required: adding the ability to create BNA animation files, figuring out how to appropriately splice image frames into the format used by the game (if at all)

### LINKED ISSUE #101: TAID0 map has improper coordinates for placing map characters
All of the other maps are correct and the pathing map in the map editor is correct, so this is a bit ??? for me.

Pathing map:
image

The farthest Kyon can be dragged left/right:
image

jonko0493:
Currently lacks any UI, but it works (with caveats)!

Firstly, fixes #101 as that was just a result of TAID0 (the gym map) happening to be a puzzle phase map! Weird since it's not labeled like the others.

Secondly, closes #89, finally! With some caveats. Notably:
1. There is no actual UI for this yet
2. ~As a result, you can't set the position of the eye and mouth animations! They're currently hardcoded to random values to prove that they work, but will need UI to actually implement. We also might consider embedding the positions in the body's filename (like we do timings).~ Fixed, you can now do this through filename.
3. ~Currently, it doesn't save changes to CHRDATA.S, so changes to the character and marking it as large or not don't propagate currently. I'll fix this before merge.~ Fixed.


jonko0493:
There was a problem where swapping between large/small sprites wouldn't work. That has been fixed with a lot of RE work into figuring out how textures get allocated!

WiIIiam278 (approved):
Looks good now :)

PR #359, opened 2024-07-17
Fixup loading SFX & PALEFFECT preview

jonko0493:
So basically, I was frustrated trying to add an SFX to a scene and realized through debugging that my assumption that SFX are all loaded in at once via the SDAT groups on a per-scene basis was wrong. In actuality, here is the "play sound" chunk of the `SND_PLAY` command's routine:

```c
if ( startStopMode == SFX_START )
{
  crossfadeDupe = parameters->crossfadeDupe;
  loadSoundIfLessThanZero = crossfadeDupe < 0;
  if ( crossfadeDupe < 0 )
    loadSoundIfLessThanZero = parameters->crossfadeTime < 0;
  if ( loadSoundIfLessThanZero )
  {
    snd_loadSound(parameters->soundIndex);
    volume = parameters->volume;
    if ( volume < 100 )
      snd_play(parameters->soundIndex, volume, 6);
  }
  else
  {
    sndIndex = parameters->soundIndex;
    volume_1 = parameters->volume;
    if ( crossfadeDupe < 0 )
      snd_play(sndIndex, volume_1, parameters->crossfadeTime);
    else
      snd_play(sndIndex, volume_1, crossfadeDupe);
  }
}
```

As you can see, it does a little dance with the two crossfade parameters: if _both_ of them are less than zero, it loads the sound into memory. Otherwise, it does its crossfading stuff.

This PR reflects this new information and adds a checkbox for loading new SFX in. It also makes that checkbox turned on by default when adding commands and adds a little functionality to disable the crossfade numeric stepper when it's checked.

This PR also fixes up `BG_PALEFFECT`'s preview (so that it applies to the background) and also adjusts the sepia filter to look closer to what it looks like in-game.


WiIIiam278 (approved):
Looks good, nice :)

PR #360, opened 2024-07-18
Add event table support

### LINKED ISSUE #334: Factor in EVTTBL.S when dealing with script files

jonko0493:
Closes #334.

This adds full support for the event table functionality, including allowing editing the SFX group for each script included in the event table, adding support for adding/removing read flags, adding new scripts to the event table, recalculating the read flag map, and adding the flag names for the read flags for the save file editor.


WiIIiam278 (approved):

WiIIiam278:
Also worked fine to me :)

PR #365, opened 2024-07-26
Add serializable script templates

### LINKED ISSUE #149: Make script templates serializable, load from folder
Script template objects should be serializable and readable from JSON files from a folder in a local directory. This will allow users to make their own custom templates and exchange them, which may be particularly useful on larger projects with map layouts. Perhaps an option to make/save a template _from_ a script would be nice.

jonko0493:
Closes #149.

Redesigning script templates to be serializable and additionally adding functionality to export the current script to a template! Neat.

Also, I added localization to the config stuff so that early stage messages still get localized!


WiIIiam278 (approved):
Nice work, yep, nice to see this finally followed up on! :)

PR #370, opened 2024-07-27
Add project import/export

### LINKED ISSUE #248: Import/Export button for projects
~Perhaps we should rename the base project file to just be an XML and make slproj a zip file that contains the proj.xml + base directory.~ nah, let's just make a new thing called an .slzip or something.

Will also need to rewrite paths to be local rather than absolute.

We also need to scrub the original ROM files... probably should store the base ROM hash somewhere to verify that it's good on import.

jonko0493:
Closes #248.

This adds a project import/export button. They still need icons @WiIIiam278. Dead simple process, but totally works -- we just create a zip archive of the proj file + font charset + portions of the base directory. On import, we unpack a base ROM (which we check for integrity with a filehash), unzip the slzip, copy to the iterative directory, then do a build before loading the project.

Additionally, this removes the `MainDirectory` entry from the slproj format. We now construct that from config + project name. Should be reliable enough and makes the proj format portable.

### DIFF
--- a/src/SerialLoops.Lib/IO.cs
+++ b/src/SerialLoops.Lib/IO.cs
@@ -180,8 +180,17 @@ namespace SerialLoops.Lib
             }
         }
 
-        public static void CopyFiles(string sourceDirectory, string destinationDirectory, ILogger log, string filter = "*")
+        public static void CopyFiles(string sourceDirectory, string destinationDirectory, ILogger log, string filter = "*", bool recursive = false)
         {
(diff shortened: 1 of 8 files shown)


WiIIiam278 (commented):
Will push icons back to after the UI rewrite as we'll to reexport and evaluate the icon set as SVGs for that :) - but otherwise, tested & looks good

WiIIiam278 (approved):

PR #373, opened 2024-07-30
Begin the Avalonia migration

jonko0493:
Yeah, it's a big one.

This deletes the entire (!) UI project. In it's place, we're building an Avalonia application. So far, the following things have been implemented:

* Home panel
* About dialog
* Preferences dialog
* Update available dialog
* New project dialog, project creation
* Open project panel (item explorer & editor tabs)
* Background editor
* Image resize/crop dialog (kinda, not working fully)

This will all need a pass from @WiIIiam278 to make it look nice (and work right in some cases).


WiIIiam278 (approved):
Merging to unblock - have not _fully_ tested yet but gave a launch with a test project (did need to clean my repo out lol). Nice work so far. Icons specifically are a bit of a point to look at.

Can begin by merging this into the Avalonia branch and do continued work bit-by-bit?

PR #376, opened 2024-07-31
Add a font setting option

Xzonn:
Use different fonts by default based on the language, and allow users to choose the font they want to use. It seems to run normally on Windows in Simplified Chinese, Japanese and English.

<details>
<summary>Screenshots</summary>

English
Simplified Chinese
Japanese
</details>

### DIFF
--- a/src/SerialLoops/ViewModels/Dialogs/PreferencesDialogViewModel.cs
+++ b/src/SerialLoops/ViewModels/Dialogs/PreferencesDialogViewModel.cs
@@ -110,7 +110,7 @@ namespace SerialLoops.ViewModels.Dialogs
                     },
                     new ComboBoxOption([
                             ("", string.Format(Strings.Default_Font_Display, Strings.Default_Font)),
-                            ..SystemFonts.Collection.Families.Select(_ => (_.Name, _.Name)).ToList(),
+                            ..SystemFonts.Collection.Families.Select(_ => (_.Name, _.Name)),
                         ])
                     {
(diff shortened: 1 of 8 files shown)


src/SerialLoops/ViewModels/Dialogs/PreferencesDialogViewModel.cs
```diff
@@ -106,6 +108,19 @@ public void Initialize(PreferencesDialog preferencesDialog, Config config, ILogg
                             RequireRestart = true;
                         },
                     },
+                    new ComboBoxOption([
+                            ("", string.Format(Strings.Default_Font_Display, Strings.Default_Font)),
+                            ..SystemFonts.Collection.Families.Select(_ => (_.Name, _.Name)).ToList(),
```
jonko0493, line 113:
```suggestion
                            ..SystemFonts.Collection.Families.Select(_ => (_.Name, _.Name)),
```
collection expression here means we can use the `IEnumerable` without converting to a `List`, which optimizes slightly

jonko0493 (approved):
With the one nit, this looks great. Especially clever thinking with loading the default font through the localized strings!! Great work all around.

WiIIiam278 (approved):
Approving to unblock bar jonko's nit, nice stuff!

jonko0493:
Thinking about this more, this is also cool because it will let people pick, like, dyslexia-friendly fonts and other accessibility-related things! Very neat.

PR #378, opened 2024-08-01
Fixup the crop/resize canvas scaling bug

jonko0493:
I finally found the issue! Weirdly, for some reason, the scale parameter was getting set to 1.5. Resetting it to 1 every time we paint fixes the issue.


WiIIiam278 (approved):

jonko0493:
The reason the scale was at 1.5 was because my monitors are scaled to 150%. Wow.

PR #381, opened 2024-08-01
Add headless Avalonia unit tests

jonko0493:
This add some basic unit test coverage for the Avalonia port:
* Tests for the ProjectCreationDialog and project creation generally
* Test for the background editor (verifying that items open correctly)
* Test for the BGM editor (verifying that items open correctly and that playback works to some extent)

In the future, we should add support for basic testing of newly added elements as they go.

Future work/maybe I will update this PR:
* I'm going to add a test for the systex editor (possibly just making the BG editor one a theory)
* I'm going to add a test for the BGM extras data -- I feel like that might depend on the below bullet point
* I may make the unit tests more unit-y and thus more parallelizable by having each unit test create its own project.

Additionally, this PR fixes the `LoopyLogger` to be less fussy about multiple instances of SL running by removing the dependency on having an open `StreamReader` and instead having retry-hardened logic for appending text to the file on a log-by-log basis. Finally, this PR bumps us to an internal build of Tabalonia that has a light theme now.


WiIIiam278 (approved):
seems nice to me!

jonko0493:
Unfortunately, had to remove the audio play tests. They worked great locally, but only ran successfully on mac in azure pipelines (probably bc those agents are physical machines with sound hardware). We can look into testing it on mac & windows at least for the UI tests.

PR #382, opened 2024-08-01
Add project settings dialogue, improve preferences

WiIIiam278:


src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
```diff
@@ -0,0 +1,119 @@
...
+using ReactiveUI;
+using SerialLoops.Assets;
+using SerialLoops.Lib;
+using SerialLoops.Lib.Util;
+using SerialLoops.Views.Dialogs;
+using SkiaSharp;
+
+namespace SerialLoops.ViewModels.Dialogs
+{
+    public class ProjectSettingsDialogViewModel : ReactiveObject
```
jonko0493, line 17: As mentioned previously, would recommend changing this to `ViewModelBase` and using the CommunityToolkit helper methods instead of the ReactiveUI ones -- if we're going to go with Reactive tho we should go all in. Up for discussion.

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
```diff
@@ -0,0 +1,119 @@
...
+            get => _gameTitle;
+            set => this.RaiseAndSetIfChanged(ref _gameTitle, value);
+        }
+
+        private SKBitmap _icon;
+        public SKBitmap Icon
+        {
+            get
+            {
+                SKBitmap preview = new(64, 64);
```
jonko0493, line 40: I assume the reason for this is that the icons are very small by default and this makes them easier to view? If so, clever!

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
```diff
@@ -0,0 +1,119 @@
...
+            ApplyCommand = ReactiveCommand.Create(ApplyCommand_Executed);
+            CancelCommand = ReactiveCommand.Create(_settingsDialog.Close);
+        }
+
+        private async void ReplaceCommand_Executed()
+        {
+            FilePickerOpenOptions options = new()
+            {
+                Title = Strings.Open_ROM,
+                FileTypeFilter = [new FilePickerFileType(Strings.Chokuretsu_ROM) { Patterns = [ "*.png", ".jpg", "*.jpeg", "*.bmp", "*.gif" ] }]
```
jonko0493, line 71: Do you mind refactoring these patterns for supported images into one of our Shared or Helper classes and referencing it from there? Also, the string you have here is `Chokuretsu_ROM` but it probably should be whatever the "Supported image formats" string is

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
```diff
@@ -0,0 +1,119 @@
...
+            };
+            IStorageFile image = (await _settingsDialog.StorageProvider.OpenFilePickerAsync(options)).FirstOrDefault();
+            if (image is null)
+            {
+                return;
+            }
+            SKBitmap fileImage = SKBitmap.Decode(image.TryGetLocalPath());
+            if (fileImage is null)
+            {
+                await MessageBoxManager.GetMessageBoxStandard(Strings.Error, Strings.Invalid_image_file_selected, ButtonEnum.Ok, MsBox.Avalonia.Enums.Icon.Error).ShowWindowDialogAsync(_settingsDialog);
```
jonko0493, line 81: Probably should just do `Log.LogError` here instead?

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        private async void ReplaceCommand_Executed()
+        {
+            FilePickerOpenOptions options = new()
+            {
+                Title = Strings.Open_ROM,
+                FileTypeFilter = [new FilePickerFileType(Strings.Chokuretsu_ROM) { Patterns = [ "*.png", ".jpg", "*.jpeg", "*.bmp", "*.gif" ] }]
+            };
+            IStorageFile image = (await _settingsDialog.StorageProvider.OpenFilePickerAsync(options)).FirstOrDefault();
+            if (image is null)
```
jonko0493, line 74: I think we should stick with the pattern here and wrap the rest of the method in an if statement of `if (image is not null)`. but that's a nit for sure, feel free to ignore if you disagree strongly

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
```diff
@@ -0,0 +1,119 @@
...
+            SKBitmap newIcon = new(32, 32);
+            fileImage.ScalePixels(newIcon, SKFilterQuality.High);
+            Icon = newIcon;
+        }
+
+        private async void ApplyCommand_Executed()
+        {
+            if (_gameTitle.Length is < 1 or > 127)
+            {
+                await MessageBoxManager.GetMessageBoxStandard(Strings.Error, Strings.Please_enter_a_game_name_for_the_banner__between_1_and_128_characters_, ButtonEnum.Ok, MsBox.Avalonia.Enums.Icon.Error).ShowWindowDialogAsync(_settingsDialog);
```
jonko0493, line 94: Consider using `Log.LogError`

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
```diff
@@ -0,0 +1,119 @@
...
+                return;
+            }
+
+            if (_gameTitle.Split('\n').Length > 3)
+            {
+                await MessageBoxManager.GetMessageBoxStandard(Strings.Error, Strings.Game_banner_can_only_contain_up_to_three_lines_, ButtonEnum.Ok, MsBox.Avalonia.Enums.Icon.Error).ShowWindowDialogAsync(_settingsDialog);
```
jonko0493, line 100: Consider using `Log.LogError`

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (Icon is not null)
+            {
+                _settings.Icon = _icon;
+            }
+
+            Applied = true;
+            _settingsDialog.Close();
+        }
+
+        private static (string Culture, string Language) GetLanguageComboxBoxOption(string culture)
```
jonko0493, line 114: Is this here by accident? Or are you planning to use it in the future?

src/SerialLoops/ViewModels/MainWindowViewModel.cs
```diff
@@ -274,6 +275,15 @@ public async Task<bool> CloseProject_Executed(WindowClosingEventArgs e)
             return cancel;
         }
 
+        public async void ProjectSettingsCommand_Executed()
```
jonko0493, line 278: Should be `async Task` rather than void.

src/SerialLoops/ViewModels/MainWindowViewModel.cs
```diff
@@ -119,6 +119,7 @@ public MainWindowViewModel()
 
             SaveProjectCommand = ReactiveCommand.Create(SaveProject_Executed);
 
+            ProjectSettingsCommand = ReactiveCommand.Create(ProjectSettingsCommand_Executed);
```
jonko0493, line 122: Since the method it's calling is async, you should use `ReactiveCommand.CreateFromTask` instead.

src/SerialLoops/Views/Dialogs/PreferencesDialog.axaml
```diff
@@ -15,16 +15,28 @@
         Icon="/Assets/serial-loops.ico"
         Title="{x:Static assets:Strings.Preferences}">
     <DockPanel>
-        <ScrollViewer DockPanel.Dock="Top" VerticalScrollBarVisibility="Visible" Height="{Binding ScrollViewerHeight}">
-            <StackPanel>
-                <controls:OptionsGroup Name="BuildOptions" />
-                <controls:OptionsGroup Name="ProjectOptions" />
-                <controls:OptionsGroup Name="SerialLoopsOptions" />                
+        <StackPanel Spacing="5" Margin="5">
```
jonko0493, line 18: For my own curiosity, is there a defined behavior when we don't specify where to dock the control in the dock panel?

jonko0493 (commented):
Few things to change before I hit approve, but great work. I love the added custom controls and love love love you redoing the UI as always. Excellent work!

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
jonko0493: If you don't wanna do this you don't have to btw lol

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
WiIIiam278: i'm kinda a guard clause fan tbh but agree w/ consistency, i'll wrap the stuff in an `if` for now

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
WiIIiam278: uhhh maybe in another PR but yeah we should do this.

actually we should also do that for Message Boxes because the syntax is very verbose now lol

src/SerialLoops/ViewModels/Dialogs/ProjectSettingsDialogViewModel.cs
WiIIiam278: nah a copy/paste LO, good catch

src/SerialLoops/Views/Dialogs/PreferencesDialog.axaml
WiIIiam278: yeah it defaults to "Left" (enum 0)

jonko0493 (approved):
👍 :shipit:

PR #383, opened 2024-08-01
Cleanup, cleanup, everybody do your share

jonko0493:
This PR does some much needed cleanup!

* Removes the dependency on CommunityToolkit for MVVM and uses ReactiveUI instead. We were already using ReactiveUI to do those fun commands, so this helps consolidate our stack a bit. It also has some very nice attributes you can apply to your properties to autogen reactivity/fields for them. Very slick.
* Refactors the MessageBoxManager into a single extension method that can be called on `Window`s. It spins up messagebox params to make them look mae yori motto oshare.
* Refactors the StorageProvider calls to the file picker into a set of four methods extensions that can also be called on `Window`s. This makes raising a file picker easier as there are fewer things to new up and keep track of.
* Refactors commonly used filetype filters into the `Shared` class


WiIIiam278 (approved):
love me a PR that removes more lines than it adds

PR #388, opened 2024-08-03
Switch to more reliable image display control

jonko0493:
Switching to an Image source that doesn't have scaling issues and will reliably convert our images. Yay!


jonko0493:
@FuyukoAyumu is blocked on this being merged. Merging without review because I'm confident in these changes.

PR #390, opened 2024-08-04
Add ASM hacks dialog

jonko0493:
Adds back the ASM hacks dialog. Fully functional with some added bugfixes. ~Tests incoming.~

While I was working on it, I also fixed the tab offsets so they're consistent cross-plat.


jonko0493:
We don't have docker on windows and we don't have devkitARM on Mac/Linux so we can't actually test patch application lol. This tests some of the dialog stuff and that the cancel button works tho ig?

Dazawassa (approved):
Parameters box doesn't appear to do anything currently. Everything else appears functional. I wasn't able to force any crashes or unexpected behaviour so it all seems to be stable and good for a merge.

jonko0493:
@Dazawassa did you select both hacks? The first hack has no parameters but the second does, so if the parameters box isn't doing anything on your box that's a problem.

WiIIiam278 (approved):
Tested & good - lets get this merged up

PR #393, opened 2024-08-26
Add back voiced line editor

jonko0493:
What it says on the tin; looks pretty great i think!

Can't add a test for this one as the sound engine can't be opened on the CI machines (same as BGM editor).


WiIIiam278 (approved):
Only partially tested this one but worked for what I tried to do. Good stuff

PR #395, opened 2024-09-12
Switch from TreeView to TreeDataGridView for ItemsListPanel

jonko0493:
TreeView is known to be very slow as it doesn't virtualize properly (and indeed, searching was kind of laggy as was expanding and collapsing categories at times). This switches to the much more optimized TreeDataGridView. There's a few tricks I use here that I'm not totally happy with and we might consider changing down the line (looking at you @WiIIiam278 😊).

For now, though, it works nicely.


jonko0493:
oh right all the tests rely on treeview stuff to operate lol. i'll fix those in a bit.

jonko0493:
Tests fixed now, yay!

WiIIiam278 (approved):

PR #402, opened 2024-11-04
Add KBG replacement, true localization for SerialLoops.Lib, and fixup LoopyLogger error logging on Avalonia

### LINKED ISSUE #225: Add KBG export and replacement
Kinda forgot about these, but we should do them!

jonko0493:
Closes #225.

The core of the work was originally adding KBG replacement, which this does accomplish. There were quite a few Chokuretsu lib changes along the way, as expected, but it should all be working now.

I've also been bothered for a while that we don't have localization in the Library project in the Avalonia world. I found out that if we generate the resxs using the "Public" generator it makes the `ResourceManager` object public, which means we can pass the manager to the library to localize strings without static access! Neat!

And then while I was there, I realized there was a bug in our implementation of the LoopyLogger with its calling of message boxes asynchronously. We now use the Dispatcher to do that call appropriately, making the message boxes pop up as expected, even when called from inside the lib (or Chokuretsu lib, even, though that is something we should avoid generally).


WiIIiam278 (approved):
Ah the lib localization is a nice idea. Cool. KGB replacement also works nicely I think :)

PR #406, opened 2024-11-08
Switch to file-scoped namespaces

jonko0493:
No change except switching to the newer-style "file-scoped namespace" for everything in the solution.


WiIIiam278 (approved):
oh I'm a big fan, this basically deindents everything by one which is MUCH nicer for working with

although merging this will create 9 trillion merge conflicts so squash this first?

PR #412, opened 2024-11-09
Add Serial Loops installers & packages generated by official build (but don't let them release yet)

### LINKED ISSUE #132: Add MSI
We should create MSI artifacts as part of our releases to better support Windows.

jonko0493:
This PR adds a ~pipeline~ GitHub action that generates the following artifacts (***bolded*** ones are recommended):
* Linux
  - ***Flatpak*** -- comes with sandboxed devkitARM; adding make and OpenAL next
  - Debian dpkg -- installs make and OpenAL alongside it
  - Red Hat RPM -- installs make and OpenAL alongside it
  - Tarball -- no dependencies
* macOS
  - ***Pkg installers for both ARM and x64*** -- installs devkitARM and calls `xcode-select --install` to install make
* Windows
  - ***Installer*** -- installs devkitARM and Docker Desktop, optionally (closes #132)
  - Portable zip -- no dependencies

Additionally, there is a small change to the ConfigFactory that checks for the flatpak location of devkitARM.

~This PR is blocked on #409 as attempting to write to protected directories (which is both where installers want to install & the case in the Flatpak's sandbox) causes SL to crash.~

A test release with all these artifacts can be seen [here](https://github.com/jonko0493/SerialLoops/releases/tag/0.0.1).


WiIIiam278 (approved):
amazing, tested and it just.. works. very significant update indeed.

WiIIiam278:
my sole comment would be labelling these file names a touch more clearly in the release artifacts
like the windows installer should prob have installer in it, the zip "portable" in it, etc. otherwise perfect

.github/workflows/official-build.yml
```diff
@@ -0,0 +1,300 @@
...
+      - name: Upload Windows zip
+        uses: actions/upload-artifact@v4
+        with:
+          name: windows-x64-zip
+          path: src/SerialLoops/bin/Release/net8.0-windows/win-x64/SerialLoops-winx64.zip
+          retention-days: 1
+      - name: Download ISCC, Docker Desktop Installer, and devkitPro Updater
+        shell: pwsh
+        run: |
+          ((Get-Content -Path install\windows\serial-loops.iss) -replace '#VERSION#', "$env:SLVersion") | Set-Content -Path install\windows\serial-loops.iss
```
WiIIiam278, line 200: image
I notice this version placeholder is still present - possible oversight? or just an effect of the test workflow?

jonko0493: Should have been fixed now -- if this is from the 0.0.1 version, it wasn't fixed then!

PR #420, opened 2024-11-12
Turn on nullable and remove as many warnings as possible

jonko0493:
I apologize in advance for this one. 161 files changed, jesus.

We have like 50 warnings in the build currently about using nullables in a non-nullable context. So I flipped on nullable context (because it is useful!) and went through our code to fix up every nullable warning I could. :)


jonko0493:
I am aware this will involve many merge conflicts as other PRs get merged. That's fine. I'm okay with that. :")

jonko0493:
Also if you think this sucks, it's fine if we remove it lol.

jonko0493:
This causes too many bugs... I think I'm gonna abandon it, despite it taking me hours to do lol.

PR #429, opened 2024-12-16
Make the search dialog a bit more MVVM-y

jonko0493:
Does what it says on the tin! I think it looks pretty good but lemme know if you have thoughts.


WiIIiam278 (approved):
Great - and thanks for doing this, is good for my own reference above all else :)

Issue #353, opened 2024-07-12
Add ability to change source image in layouts


jonko0493:
In the process of completing #259, it was discovered that changing the `RelativeShtxIndex` parameter in the layouts does not change the image in the game. It's currently unknown _why_ this is the case. Figuring out how to change these parameters isn't mission critical for layout editing, but it might have consequences further down the line (#89 and #100 come to mind).

PR #430, opened 2024-12-19
Add layout editor to Avalonia branch

### LINKED ISSUE #352: Improve layout editor UX
I designed it, so it needs to be fixed. :)

jonko0493:
This ports over the layout editor to the Avalonia branch and it's more functional than last time!

I honestly think this is pretty good and maybe closes #352, with your approval @WiIIiam278. However, due to #353 still being open, this PR does not have a button to modify the relative SHTX index (as we've seen that doesn't seem to do anything, strangely). When we figure that out in the future, I'll add it.

@FuyukoAyumu, use this branch (or wait until it merges) to produce a JSON version of the layout you created and then check that into the assets repo, please.


WiIIiam278 (approved):
Very nice :)

PR #431, opened 2024-12-19
Add more script commands & save editor

jonko0493:
This PR adds the following script command editors:

* `BG_FADE`
* `TRANS_OUT`/`TRANS_IN`
* `SET_PLACE`
* `ITEM_DISPIMG`

It also reimplements the save editor in Avalonia. I opted to make the save editor an editor rather than a standalone dialog -- this makes sense for integrating it with the project, after all, and limits the number of open dialogs we have at any given time. New features have also been added to the save editor, notably the ability to edit the most recent objective allocations and the character power statuses. ~Broken out into separate PR~ Reverted to this PR to avoid merge conflicts.

Additionally, this PR introduces some bugfixes to previous script editor commands and uses KBG04 in `OP_MODE` preview.


WiIIiam278:
I disagree about making the save editor an item editor in the project. While I see the benefits in terms of the code structure, I think it's super important actually we keep this functionality separate, and look at making it not actually require an open project in the future as we originally discussed.

WiIIiam278 (approved):
Nice work on the new item commands :)

I disagree - fairly strongly - about making the save editor an item editor in the project. While I see the benefits in terms of the code structure, I think it's super important actually we keep this functionality separate, and look at making it not actually require an open project in the future as we originally discussed.

This PR does two very different things I think - would you mind removing the save editor stuff from this and moving it to another branch for the time being? I feel quite strongly about the save editor and think we should have a chat about how we design it when moving it to Avalonia before implementing :).

WiIIiam278 (changes requested):
whoops I did NOT mean to approve lol, that was close - lets chat first :)

jonko0493:
Yep, will cherry pick the save editor commits to a different PR; sorry about combining them in one haha. I think the benefits of having the save editor be an editor outweigh the cons, and I will explain myself!!

WiIIiam278 (approved):

PR #432, opened 2024-12-19
Add references dialog, small improvements to script editor

### LINKED ISSUE #414: When opening an .slproj, open the clicked project after the app initializes
Need to handle the slproj filepath arg

jonko0493:
This PR adds:
* The references dialog back (right click item explorer -> Find References...) 
* Some small tweaks to the script editor &ndash; notably, it adds a GridSplitter that allows for resizing the commands sidebar and reduces the length of conditionals in `VGOTO` command lines.
* The "Font" option combo box in the preferences dialog now displays each font option in its font 
* Item renaming (super simple stuff!)
* Double click a .slproj file to open it! Closes #414 
* Fixes up the updater dialog (no longer crashes, markdown works now!)
* Adds a crash logger


src/SerialLoops/Models/ItemDescriptionTreeItem.cs
```diff
@@ -1,44 +1,104 @@
...
         this.OneWayBind(ViewModel, vm => vm.DisplayName, v => v._textBlock.Text);
         this.Bind(ViewModel, vm => vm.DisplayName, v => v.Text);
         _panel.Children.Add(_textBlock);
+
+        if (_window is not null)
+        {
+            ICommand findReferencesCommand = ReactiveCommand.Create(FindReferences);
+            _panel.ContextMenu = new();
+            MenuItem references = new() { Header = Strings.Find_References___, Command = findReferencesCommand};
+            _panel.ContextMenu.Items.Add(references);
```
WiIIiam278, line 53: No idea how to do this as we'd have to somehow pass the cell and programatically trigger the data grid edit... but we should 100% have the Rename command be a context menu item here :). Maybe we have an "Item settings" dialog with more item-specific options (e.g. perhaps settings for chess puzzles, scripts), including rename?

WiIIiam278 (approved):

WiIIiam278:
Good stuff, approved - did leave a comment about a future thing to look at.

Another comment I forgot to add: With the item panel sidebar resizing -- this is nice! But not particularly useful (?), since the command text preview is limited. It might be more useful if we made that a bit longer and had the UI stylistically truncate the text with an ellipse.

Also, we should 
- remove the column header from the command table column
- prevent the table resize (as technically we now have double resize
- fix this to the width of its newly resizable parent,
- set a minimum width,
- and left-align the edit action buttons.
 
Will PR this :)

jonko0493:
Good luck on your UI changes 🫡🫡🫡

src/SerialLoops/Models/ItemDescriptionTreeItem.cs
jonko0493: I'm not sure about an item settings dialog, but that would be an elegant solution if we could come up with more than just renaming things in there. Maybe there's a way to manually trigger rename?

WiIIiam278:
> Good luck on your UI changes 🫡🫡🫡

Think I'm ahead of you here 😎

PR #433, opened 2024-12-20
A few small style improvements

WiIIiam278:
* Improves resizing of item command column 
* Improves toolbar styling (I wasn't a big fan of the accented BG on the button items - now it's all flush with the main BG)
* Fixes an incredibly subtle issue with the centering of text in tree items.... everything lines up nicely now!

### DIFF
--- a/src/SerialLoops/App.axaml
+++ b/src/SerialLoops/App.axaml
@@ -35,6 +35,7 @@
 
         <!-- Toolbar styling -->
         <Style Selector="Button.ToolbarButton">
+            <!-- We use toolbar height here so these buttons are always proportional & square in the bar -->
             <Setter Property="Width" Value="{DynamicResource ToolbarNormalHeight}" />
             <Setter Property="Height" Value="{DynamicResource ToolbarNormalHeight}" />
             <Setter Property="Padding" Value="5" />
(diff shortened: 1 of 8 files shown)


jonko0493 (approved):
Fantastic work here 😎😎😎

jonko0493:
Ah, one small note: at least on my machine, this is happening with the toolbar now:

image

src/SerialLoops/App.axaml
```diff
@@ -32,6 +32,23 @@
         <mintoolbarthemes:MiniToolbarStyles/>
         <StyleInclude Source="avares://Avalonia.Controls.ColorPicker/Themes/Fluent/Fluent.xaml"/>
         <StyleInclude Source="avares://Avalonia.Controls.TreeDataGrid/Themes/Fluent.axaml"/>
+
+        <!-- Toolbar styling -->
+        <Style Selector="Button.ToolbarButton">
+            <Setter Property="Width" Value="{DynamicResource ToolbarNormalHeight}" />
```
jonko0493, line 38: Where are these properties defined, btw?

jonko0493: I think this might be the issue

WiIIiam278: > Where are these properties defined, btw?

https://github.com/profix898/MiniToolbar.Avalonia/blob/f2628e8c8c1bd6bf8fa0680b94ed4b870e868496/MiniToolbar.Avalonia/Themes/MiniToolbarStyles.axaml#L4

windows issue as worked on mac, will take a look

jonko0493: ah, that makes sense yeah. I will say you defined the width with "ToolbarNormalHeight" -- not sure if that's intentional, but if it is it could use a comment

PR #434, opened 2024-12-20
Add save editor lol

jonko0493:
This PR reimplements the save editor in Avalonia. I opted to make the save editor an editor rather than a standalone dialog -- this makes sense for integrating it with the project, after all, and limits the number of open dialogs we have at any given time. New features have also been added to the save editor, notably the ability to edit the most recent objective allocations and the character power statuses.

It was originally part of https://github.com/haroohie-club/SerialLoops/pull/431, but I branched it off. There will be horrendous merge conflicts as a result that I will bear the brunt of. lol


jonko0493:
Actually, splitting this sucks so I'm not gonna do it 😎

PR #437, opened 2024-12-20
Make the save editor work as a standalone feature as well

### LINKED ISSUE #369: Separate "Edit Save File" mode from "Edit Project" mode
When opening a save file from the home screen, we should _not_ actually create a new project like we currently do. Rather, we should create a temp project in read-only mode. Editing a save file from an open project should still be possible but should perhaps come with a warning.

jonko0493:
Now, from the home screen, you can select "Edit Save File" and be prompted to select a save file. Once you do, a project will automatically be created from either a ROM with the same name as the save or a ROM you select if no such ROM exists. It then loads the save editor as the main content of the screen.

Technically, this closes #369.


WiIIiam278 (approved):
Very cool, works very nicely indeed!

PR #438, opened 2024-12-21
Add the map editor to the Avalonia branch

jonko0493:
Comes with the brand new ability to filter the map by layers & can now display the scrolling backgrounds in the puzzle phase. It doesn't take super long to load and once it does filtering is blazingly fast because we don't need to re-render the graphic each time an option is changed! Very cool.

Also comes with a bunch of new RE discoveries, including:
* The existence of the aforementioned layers (the main ones are BG, Occlusion, and Object layers, where the BG and Occlusion layers each pull from the "BG object" layer as well.)
* The existence of "junk" layers (BG and object) that include the entries that previously were mucking up the map previews -- all maps now render pixel perfectly
* Additional data layers, including the origin (pre-existing but now can be visualized), bounds, and some other extra ones
* The existence of a transform mode for scrolling BGs which defines how the scrolling BG elements should be rotated/flipped (not yet implemented bc it made me sad)
* The existence of (unused!) camera trucking definitions that play when the map is initially loaded


jonko0493:
Drafting this for now as I'm making some good RE progress on these and want to see how much I can figure out before we merge.

jonko0493:
There's cool rotation stuff happening with the scrolling BGs. I don't want to deal with it rn, so I'm not gonna. :D

jonko0493:
@WiIIiam278, undrafted and ready for review for now I think. We can do more work on it when we get closer to the milestone where we're actually gonna edit these.

WiIIiam278 (approved):
great now the pan n' zoom works!

PR #439, opened 2024-12-23
Add more search scopes

jonko0493:
Adds three new search scopes:

* *Command* -- search script commands including a pretty powerful syntax for searching with parameter filters
* *Archive Index* -- search by an item's corresponding file's or (files') index in its original archive
* *Archive Filename* -- search by an item's corresponding file's (or files') name in its original archive


WiIIiam278 (approved):
Cool! We deffo need some tooltips or UX for the command scopes. I also think perhaps we should have some sort of expand/collapse thing for the search panel as there are a LOT of options now. Cool, though!

PR #441, opened 2024-12-24
Add hack creation dialog, debug printing hack

### LINKED ISSUE #440: ASM hack creation dialog
Would be useful to have a system for creating ASM hacks. Not an automated programming interface, ofc, but rather something for creating the serialized file format that we use for them.

jonko0493:
This adds a dialog for creating ASM hacks (closes #440). You give it hack files and then it does a bunch of parsing and such to auto-generate the serialized hack JSON, including parsing to find injection sites, undefined symbols, and parameters. It then prompts the user to define these values for each hack file. It can also detect the difference between hook files and repl files and even handle C files.

Additionally, this adds the debug printing hack. This hack is already applied to English base ROMs but having it here means it can now be applied to the Japanese base ROM as well! Neat.


WiIIiam278 (approved):

WiIIiam278 (approved):

WiIIiam278:
whoops I approved this twice somehow??

PR #460, opened 2025-02-14
MVVM-ify progress dialog, fixup progress dialogs on Linux

jonko0493:
Will's fantastic work to make the app look stylish made it look not as good on Linux where acrylic blur isn't as available. This was because the progress dialog was difficult to make look good when it didn't have XAML. Now it does!


WiIIiam278 (approved):
Nifty bit of housekeeping, thanks :)

PR #463, opened 2025-02-17
Implement lots of styling fixes, cleanup warnings

### LINKED ISSUE #462: GroupBoxClassic theme changes fonts of controls inside
We should switch to using a styled [HeaderedContentControl](https://docs.avaloniaui.net/docs/tutorials/groupbox) per this documentation.

jonko0493:
Ignore the branch name, this is a fixup PR now. :P

This PR accomplishes a variety of things that needed to be done:
1. Introduces better theming control to Linux by conditionalizing the acrylic border on platform -- the acrylic border now lives in a single control and so can be updated more easily across all dialogs that use it
2. Implements the acrylic border theming across all the dialogs in the application. A few dialogs even had their chrome removed, leading to extremely clean looks:
image
image
image
3. Fixes #462 by removing the dependency on GroupBox.Avalonia in favor of using custom-styled HeaderedContentControls.
4. Fixes a bunch of warnings the build had -- we're down to two now (and one is just an unimplemented async method that will go away when it gets implemented). This makes for a much cleaner build experience and makes it easier to tell if a warning that pops is actually valid/important.
5. The shortcuts displayed on the empty tab project space are now conditionalized by platform (so they say `Command+S` on Mac but `Ctrl+S` on Win/Linux).


WiIIiam278:
holyyy crap this is big

very cool thank you

WiIIiam278 (approved):
had a very brief scan through - we'll want to test some of the dialogs after the chrome removal + acrylic -- but very cool overall and some nice housekeeping :)

PR #466, opened 2025-02-19
Make flatpak sandbox patch & devkitARM pull more maintainable

### LINKED ISSUE #461: Extract sandboxed constants out into separate class for flatpak patch
The flatpak build patches the application to run better in a sandbox; however, these patches are fragile and require updating every time the source files are updated. We should abstract out the values that need to be patched into a static class that will require minimal changes.

jonko0493:
When we build the flatpak, we do some simple source patching to ensure that features like running the emulator and running devkitARM still work from within the flatpak sandbox. However, source patching has the flaw that your patch has to be updated whenever the source file is updated.

This PR obviates that by abstracting the values patched by the sandbox patch into a single static class that can be patched instead. This makes things much more maintainable going forward.

Additionally, this makes downloading devkitARM a bit simpler -- instead of pulling from a hosted archive, we simply pull the latest devkitARM docker container in the setup step, package up `/opt/devkitpro/devkitARM`, upload it as an artifact, download it as an artifact, and then untar that artifact as part of the build.

Closes #461.

### DIFF
--- a/.github/workflows/official-build.yml
+++ b/.github/workflows/official-build.yml
@@ -49,7 +49,7 @@ jobs:
           image: devkitpro/devkitarm:latest
           options: -v ${{ github.workspace }}:/work -w /work
           run: tar -czvf devkitarm.tar.gz /opt/devkitpro/devkitARM
-      - name: Upload devkitARM zip
+      - name: Publish devkitARM tarball
         uses: actions/upload-artifact@v4
         with:
           name: devkitARM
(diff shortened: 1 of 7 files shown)


jonko0493:
Note to will: i'm headed out for a bit, don't merge unless [this](https://github.com/jonko0493/SerialLoops/actions/runs/13475998253/job/37655256641) passes

jonko0493:
Alright, it's working now!

WiIIiam278 (approved):

PR #467, opened 2025-02-20
Add AppImage release artifact

### LINKED ISSUE #465: Create Linux AppImage
We produce a Flatpak which is great but some folks prefer AppImages, so we should produce one of those, too.

jonko0493:
This PR adds an AppImage artifact to the official release build. It's a neat one-click alternative to the tarballs and seems to work pretty well!

Closes #465.


WiIIiam278 (approved):
linux stuff! cool! :)

jonko0493:
I don't really like appimages to be honest but some people do so might as well have the option. :)

PR #468, opened 2025-02-20
Quick volume normalization fix

jonko0493:
I observed a crash during project load when calculating the max volume. It seems rare enough to not worry about (for now). This fix at least keeps things from crashing.


WiIIiam278 (approved):
tbh i'm happy casting a widder try-catch-log net around sensitive audio/impl specific logic

PR #474, opened 2025-02-24
Add working project migration to Avalonia branch

### LINKED ISSUE #351: Migrate to new ROM is probably broken
The feature should be investigated. I think we probably need to new up a project or something during the migration process? Or at least manually clear a bunch of things.

jonko0493:
Fixes #351.

Project migration was broken in the Eto branch. This gets it working in Avalonia. Made some changes to make the project creation dialog reusable by project migration so that we can also allow migrating ROMs across languages as well.


WiIIiam278 (approved):
Tested this, and it works! We should deffo improve docs on what this actually does though as I'm not sure "migrate project" means what we think people think it means. To me it sort of implies "migrate from another third party tool"

jonko0493:
Definitely should document, but if the terminology is confusing we should consider adjusting it within the program as well. Let's think on it at least.

PR #476, opened 2025-02-24
Add reopening workspace functionality to Avalonia branch and improve tests

jonko0493:
This is one of the last pieces of functionality to port over from Eto. I've also added a shiny new feature which is that it now stores the specific tab you had open when you closed the workspace and reopens to that specific tab. Spiffy!

Additionally, because the tests broke on this, I took the chance to improve our tests so that when you run them locally, they auto-download test assets. As a result, I found a bug in Avalonia's NUnit package that I fixed! Yay!

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,18 +34,11 @@ jobs:
       arguments: -f $(targetFramework)
     displayName: Build project
 
-  - pwsh: |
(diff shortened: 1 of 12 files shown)


WiIIiam278 (approved):
this isn't just big jonko, it's large.

honestly the best PR ever (lol), I love the reopen project stuff we used to have and you made it better

PR #502, opened 2025-03-08
Cleanup dialogs & dropdowns, bump to Avalonia nightly, new templates, new hacks, general bugfixes

### LINKED ISSUE #240: On most dialogs, Enter should confirm and Esc should cancel
Will just make things easier.

jonko0493:
Closes #240.

This makes sure every dialog can be closed via Esc and most can be submitted via Enter through the built-in Avalonia `IsCancel` and `IsDefault` properties.

Additionally, I have bumped us to an Avalonia nightly build so the combo boxes work correctly for the user study at least. Hopefully 11.2.6 comes out in the next week or so before we release full? If not, I'm personally fine releasing with a nightly build.

Additionally, I'm going to add some new script templates in this PR because this is the current branch with the bugfixes I need to do so.

As mentioned in the last one, I also have some general bugfixes I'm applying here.

Yeah, why not, this is an ASM hack PR now, too.

Hell yeah, I cleaned up all the ComboBoxes while I was here.


jonko0493:
Okay this PR got out of hand lmao.

Let's try to summarize everything.

Features:
* The primary goal was completing #240, so almost all dialogs now respond to Esc/Enter to close/submit
* We've bumped to Avalonia nightly to get ComboBox text search working
  - With that, we've switched mostly from manually specifying text search to simply doing .ToString() on items like WPF does
* Made changes to NitroPacker to support hacks with only `arepl` files specified, meaning hacks can now be simple replacements
* Along with that, added arepl hacks as injection sites and added a check to make sure they're applied
* Made the default for voice replacement to encode audio with standard ADX codec as our AHX encoder seems to cause bugs, but left in the option to encode as AHX (with a tooltip explanation).
* Made volume normalization capable of boosting gain on tracks as well
* Added SFXs in `GROUP_ONMEM` (always loaded into memory) to the `SND_PLAY` command SFX selection menu in addition to the already present SFXs in the script's group.
* Made the overlays in the crop/resize dialog be hit test invisible so the image can now be dragged around/zoomed by clicking in the grayed out region as well!
* Bumped the choku lib to get new features as well as some fixes (notably, significantly simplified and slightly sped up operating on scripts)

Templates:
* Adds an "isometric map start/end" template for easily adding isomap start/end sections. btw we need to rethink templates a bit but that's something i'll file for a separate issue and maybe is v0.4?
* Adds a template for topic files with sections for each character

Hacks:
* Added a hack to change the monologue text color to one of the other built-in colors
* Added a hack that makes all chibis playable in the investigation phase by replacing checks for Kyon's chibi with checks for the first chibi in the map characters section
* Added a hack that moves `ITEM_DISPIMG` displayed items up a few z-layers so they display above isomap layers but still behind character sprites (not really useful except for weirdos like me who want to use that command lol)

Bugfixes:
* Applying a template didn't set the script item to have unsaved changes
* Template names with illegal path characters would cause crashes when attempting to save them. Actually this still might be a problem now that I'm looking at the commit lol, I might revisit this again
* Generate template dialog has text wrapping now as does template selection dialog
* Inserts missing conditionals on script template application. Actually this should also be changed to just insert all of the conditionals to avoid conflicts
* We need to stop localizing parameter names since we don't use them anymore but that's a fix I have in the db change; here I just changed one parameter that was crashing template application
* ASM hack generation was parsing parameters incorrectly, assuming they had to start with a `#` instead of just looking for `{{paramName}}` like hack application does
* Ensure that arepl files are deleted when hacks are unapplied
* Ensure that tiled system textures are replaced with `newSize: true` as image height _can_ change depending on how we construct the tiled image
* Made track details in the sound player panel reactive so they can be updated
* Made `BG_FADE` actually save the specified fade duration (lmao)
* Fix symbol file creation when applying ASM hacks (were missing a `;`)
* Starting chibis were being inserted on the wrong side of the section terminator so they weren't showing up lol
* Made the `FLAG` script command actually save the correct flag ID
* No longer decrement remaining script section IDs when removing a script section as it breaks commands that reference those sections
* Gave interactable objects selector a max height so it scrolls rather than pushing the script editor down as you add more objects
* Saves the project after assigning the base rom hash on project creation, ensuring it actually gets stored in the slproj file
* Excludes arm9.bin from project export, ensuring we don't include any game code in the exported zip

WiIIiam278 (approved):

PR #512, opened 2025-03-14
Add ability to change dialogue text color

jonko0493:
This PR adds the ability to change the dialogue text color through a humble dialog. There are sixteen possible dialog colors, but only seven of them are actually used in the game (I think). The ones that have hardcoded uses (default, monologue, info, and subtitle shadow/black text) all have little help icons to explain this to the user.

image

If users want to revert to base game colors at any time, they can check the provided palette.

image

Additionally, subtitles now have color selection that looks like this:

image

And the new colors work in dialogue preview:

image

Spiffy!


jonko0493:
Also, removed my monologue text color hack as it is no longer needed.

WiIIiam278 (approved):

PR #529, opened 2025-03-18
Implement user study feedback part 2

jonko0493:
Implementing some user study feedback to improve it for the next round


jonko0493:
#526 merging will fix the test failure.

PR #530, opened 2025-03-18
Make character nameplate overridable, add two more layouts

### LINKED ISSUE #523: Allow for manual replacement of character nameplates and places
Our current style of replacement is very good for users who just want to quickly rename a character or place, but we should allow people to create nameplates/places manually in a different style or just fit longer names in themselves.

jonko0493:
A neat little feature I implemented for not-so-secret reasons :)

Closes #523.


jonko0493:
#526 merging will fix the test failure.

WiIIiam278 (approved):

jonko0493:
Going to merge this after we release v0.3, I think. Feels more like a v0.4 feature and could be a bit more fleshed out in future PRs.

Issue #533, opened 2025-03-19
"Run to Here" feature


jonko0493:
This would be a very nice feature and actually not as hard to implement as it sounds. Basically, we would hijack the save file associated with the ROM (assuming that the emulator is melonDS, this would be the save alongside the ROM; perhaps we should also allow specifying a save file directory?) and edit the quick save data to point to the current script command. Then the process for jumping to that script would be getting to the menu and hitting "continue" and selecting the quicksave.

This is minimally intrusive on the ROM and doesn't require building which is neat!

jonko0493:
Could potentially even allow for a more advanced "run to here" that pops the quick save menu to allow you to edit variables ahead of time? things to think about

PR #536, opened 2025-03-19
Implement more feedback from user studies

### LINKED ISSUE #532: Drag/dropping a map character makes it appear selected but changes don't persist
Basically, after moving a map character around, it will appear that you can make changes to it like changing the chibi or script section, but you actually have to click on it again in order for it to be considered actually selected.

jonko0493:
This implements a couple of things:

Features:
* (Not from study, just wanted): Script section re-ordering, including a little tooltip if you try to swap with/drag SCRIPT00 saying it can't be moved
* No pop-up for devkitARM not installed -- we don't do the pop-up for anything else and it's very annoying and confusing if you get in the situation where you don't have devkitARM in the standard location.
* Docker turned on by default on Windows -- thought we had this but apparently we didn't? lmao
* Grid splitters for the item explorer panel and in the script editor -- Item explorer panel can now be adjusted with a grid splitter, as can the split between the preview/tabs and command editors in the script editor. Neat.
* No resizing on some dialogs -- the BGM properties dialogs and the command add dialogs have had the ability to resize them removed so they keep their normal appearance
* Tip for type to search -- on the add command dialogs (scenario & script), there's now a little tip at the bottom advising the user they can start typing to jump to an item in the dropdown
* Tooltips on the script command editor buttons -- thought we already had these but alas
* Start maximized -- it's intended to be used maximized, so why not start them that way? ~Will check that this looks good on mac before merging lol~ per comment below, looks great
* Realized that the SELECT command's second set of parameters were not "display flags" as previously assumed but conditionals! They now parse and are handled correctly
* Tabalonia now has "Close All to Left" and "Close All but This" (now "Close All Other Tabs") has had its performance significantly improved; these context menu options are also now localizable
* Image crop/resize dialog now supports +/- for fine-grained zoom in/out and holding shift to increase the steps between zooms/pans
* Added a bunch more help tooltips throughout the editors
* Added a link to the selected map in the LOAD_ISOMAP command

Bugfixes:
* After dragging/dropping a map character, it would no longer respond to changes in the editor (appearing to be selected but not actually changing facing direction, etc.). It now reselects the map character immediately after drop. Fixes #532 
* Save editor crashed when trying to edit a blank quick save -- you're now able to create one from scratch (important for #533)
* Fixed the margin issue again for the menubar on win/linux dual monitor -- was reintroduced at some point lol
* Significant bugfix for the conditionals garbage collection process to make it actually work 😎 
* Despite having four parameters for choices, the SELECT command seems to have a hardcoded max of three, so I've removed access to changing two of the parameters.
* Fixed an issue with the export patch dialog not being reactive
* Bumped us to the new Avalonia Behaviors version because middle click to close still works as long as you bump the Tabalonia version of that package as well
* Fixed a bug with images smaller than the final image in the crop/resize dialog so that they can now be appropriately cropped/resized lol. Also shrank it so that it's easier to use on smaller resolutions
* Fixed up a very silly bug in chibi frames export

### DIFF
--- a/src/SerialLoops.Lib/Config.cs
+++ b/src/SerialLoops.Lib/Config.cs
@@ -52,10 +52,6 @@ public class Config
 
     public void ValidateConfig(Func<string, string> localize, ILogger log)
     {
-        if (string.IsNullOrWhiteSpace(DevkitArmPath))
-        {
-            log.LogError(localize("devkitARM is not detected at the default or specified install location. Please set devkitARM path."));
-        }
         if (CurrentCultureName is null)
         {
             CurrentCultureName = CultureInfo.CurrentCulture.Name;
--- a/src/SerialLoops.Lib/Factories/ConfigFactory.cs
+++ b/src/SerialLoops.Lib/Factories/ConfigFactory.cs
@@ -123,7 +123,7 @@ public class ConfigFactory : IConfigFactory
             DevkitArmPath = devkitArmDir,
             EmulatorPath = emulatorPath,
             EmulatorFlatpak = emulatorFlatpak,
-            UseDocker = false,
+            UseDocker = OperatingSystem.IsWindows(),
             DevkitArmDockerTag = "latest",
             AutoReopenLastProject = false,
             RememberProjectWorkspace = true,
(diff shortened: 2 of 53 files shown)


jonko0493:
it doesn't start in green full screen mode on macOS, just the full size of the screen, so i'd say that point is totally fine

PR #543, opened 2025-03-23
User study fixes (pt 4)

jonko0493:
A bunch of fixes/new features per requests from user studies.

Features:
* Text drawing & charset have been updated to account for the updates to the font width hack
* Ctrl-W has been rebound to close tab with Ctrl-Q being used to close the project. Ctrl-Shift-T now reopens the most recent previously closed tab (pops it off a stack of depth 20)
* Save Project and Close Project have been transplanted to the file menu, while project rename/duplicate/delete transpose themselves from the file menu to the project menu after a project has been loaded (since they have slightly different functionality between those two views, this makes a lot of sense)
* Sprite selection now happens automatically when you pick a dialogue speaker if there is a previous command with the same speaker and a non-exited sprite set -- nice little convenience factor

Fixes:
* Release text now notes that Docker Desktop requires you to _restart_ your computer as this seems more effective usually than just logging out
* Item location strings were not showing up; fixed here
* Our code for launching the emulator on macOS only worked for melonDS; it has now been genericized so it's a) more correct and b) also works with DeSmuME (important for older macs especially)
* Maps were broken by a previous change to layouts in the choku lib; this change has been reverted with some sadness
* The code to make it so you can't put spaces in project names has been fixed to work on macOS and also has been applied (even more strictly) to script sections to prevent gcc errors

Drafting for now as there is more to come but want you to be able to see what's happening so far.


WiIIiam278 (approved):
Gave a test earlier against some of the issues, nice work :)

PR #548, opened 2025-03-26
Make section edges aware of specific calling command in graph

### LINKED ISSUE #547: Command graph is wrong when commands come after the VGOTO/GOTO
If a VGOTO or GOTO appears earlier in a scene, the command graph currently includes the commands after those in the section when calculating the preview. This means that the preview can be wrong if they would normally be skipped by a VGOTO (and also a GOTO/SELECT but such a case would not happen since commands after a GOTO would never be executed).

jonko0493:
Fixes #547.

The fix for the issue is essentially to add the "source command index" (i.e. the index of the calling command) to the script section edges so that, when calculating the list of executed commands, the program knows exactly to which command the script will run before jumping to the next section. This means that commands after a VGOTO will not be included in the script preview for that section, which is exactly what we want.

### DIFF
--- a/src/SerialLoops.Lib/Items/ScriptItem.cs
+++ b/src/SerialLoops.Lib/Items/ScriptItem.cs
@@ -78,7 +78,9 @@ public class ScriptItem : Item
                     {
                         Graph.AddEdge(new()
                         {
-                            Source = section, Target = ((ScriptSectionScriptParameter)command.Parameters[4]).Section
+                            Source = section,
+                            SourceCommandIndex = command.Index,
+                            Target = ((ScriptSectionScriptParameter)command.Parameters[4]).Section,
                         });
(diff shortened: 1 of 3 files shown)


jonko0493:
@WiIIiam278 to test this, create a blank script and then apply the script intro template. Add four sections -- an INTRO section, a VGOTO section, a GOTO section, and an END section. Add an initial command to INTRO, then a VGOTO (doesn't matter what condition you choose) and point it to the VGOTO section. Add a BG_FADE to change the background, then a GOTO to the GOTO section. Finally, add a GOTO to the VGOTO section so it goes to the END section. Add DIALOGUE commands to the beginning of each of the sections and check the script previews for those. The VGOTO section will have the initial BG while the GOTO section will have the second BG set in the INTRO section.

PR #568, opened 2025-08-24
Add automated CG thumbnail replacement

jonko0493:
We just found out that CGs have separate thumbnails which are specified in the extras file. This PR displays these thumbnails in the background editor and also automates replacing them when a CG with a thumbnail is replaced.


WiIIiam278 (approved):

jonko0493:
Don't merge yet, just realized thumbnails look like ass. need smoothing

